PR #1, opened 2022-04-15
Add sct and cert verification

loosebazooka:
Add sct and cert validations

- Rename CertificateResponse to SigningCertificate
- Add more testing around parsing
- Add example fulcio/ctfe public keys and sct/cert examples for testing
- Exceptions are still kinda just passed along


loosebazooka:
@di here is working sct/cert validation. There are sample in src/test/resources. They are generated off of my email, so it does feel weird to include them in a python client, but you could get away using them to test your code for now.

src/main/java/dev/sigstore/fulcio/client/FulcioValidator.java
```diff
@@ -0,0 +1,120 @@
...
+
+  private FulcioValidator(CTLogInfo ctLogInfo, TrustAnchor fulcioRoot) {
+    this.ctLogInfo = ctLogInfo;
+    this.fulcioRoot = fulcioRoot;
+  }
+
+  public void validateSct(SigningCertificate sc)
+      throws CertificateEncodingException, FulcioValidationException {
+    if (!(sc.getLeafCertificate() instanceof X509Certificate)) {
+      throw new RuntimeException("Encountered non X509 Certificate when validating SCT");
```
vlsi, line 77: Is there a way to make the message easier to understand? E.g. `"Encountered non X509 Certificate when validating SCT. Leaf certificate class is "  + sc.getLeafCertificate().getClass()`

src/main/java/dev/sigstore/fulcio/client/FulcioValidator.java
```diff
@@ -0,0 +1,120 @@
...
+  }
+
+  public void validateCertChain(SigningCertificate sc) throws FulcioValidationException {
+    CertPathValidator cpv;
+    try {
+      cpv = CertPathValidator.getInstance("PKIX");
+    } catch (NoSuchAlgorithmException e) {
+      // no PKIX, we probably shouldn't be here, but this seems to be a system library error
+      // not a program control flow issue
+      throw new RuntimeException(e);
```
vlsi, line 96:
```suggestion
      // 
      throw new RuntimeException("No PKIX CertPathValidator, we probably shouldn't be here, but this seems to be a system library error not a program control flow issue", e);
```

src/main/java/dev/sigstore/fulcio/client/FulcioValidator.java
```diff
@@ -0,0 +1,120 @@
...
+    }
+
+    PKIXParameters pkixParams;
+    try {
+      pkixParams = new PKIXParameters(Collections.singleton(fulcioRoot));
+    } catch (InvalidAlgorithmParameterException e) {
+      // this should have been checked when generating a validator instance
+      throw new RuntimeException(e);
```
vlsi, line 104:
```suggestion
      throw new RuntimeException("Can't create PKIX parameters for fulcioRoot. This should have been checked when generating a validator instance", e);
```

build.gradle.kts
```diff
@@ -18,6 +18,7 @@ dependencies {
     implementation("com.google.api-client:google-api-client-gson:1.31.5")
 
     implementation("com.google.code.gson:gson:2.8.9")
+    implementation("org.conscrypt:conscrypt-openjdk-uber:2.5.2") // contains library code for all platforms
```
vlsi, line 21:
```suggestion
    implementation("org.conscrypt:conscrypt-openjdk-uber:2.5.2") {
        because("contains library code for all platforms")
    }
```

Adding the reason to `because` makes the reason visible to **consumers** of the library.

src/main/java/dev/sigstore/fulcio/client/FulcioValidator.java
```diff
@@ -0,0 +1,120 @@
...
+import java.util.Date;
+import org.conscrypt.ct.CTLogInfo;
+import org.conscrypt.ct.CertificateEntry;
+import org.conscrypt.ct.VerifiedSCT;
+
+public class FulcioValidator {
+  private final CTLogInfo ctLogInfo;
+  private final TrustAnchor fulcioRoot;
+
+  public static FulcioValidator NewFulcioValidator(byte[] ctfePublicKey, byte[] fulcioRoot)
```
vlsi, line 40:
```suggestion
  public static FulcioValidator newFulcioValidator(byte[] ctfePublicKey, byte[] fulcioRoot)
```

Java naming convention suggest that method names should start with a lower case letter.

loosebazooka: oh man, a short detour in go and I'm all messed up :facepalm:

src/main/java/dev/sigstore/fulcio/client/FulcioValidator.java
loosebazooka: Yeah that makes sense, I was being a little lazy with all these runtime exceptions. Lemme fix it up.

src/main/java/dev/sigstore/fulcio/client/FulcioClient.java
```diff
@@ -122,26 +115,7 @@ public CertificateResponse SigningCert(CertificateRequest cr, String bearerToken
...
-
-      byte[] certBytes = section.getBase64DecodedBytes();
-      certList.add((X509Certificate) cf.generateCertificate(new ByteArrayInputStream(certBytes)));
-    }
-    if (certList.isEmpty()) {
-      throw new IOException("no certificates were found in response from Fulcio");
-    }
-
-    return new CertificateResponse(cf.generateCertPath(certList), sct);
+    return SigningCertificate.newSigningCertificate(resp.getContent(), sctHeader);
```
vlsi, line 119: This looks like an unclosed `InputStream`, doesn't it?

loosebazooka: definitely is, lemme go fix that too.

loosebazooka: so this method signature now changed to accept string, it's probably okay to store the pem chain in memory for now.

src/main/java/dev/sigstore/fulcio/client/FulcioClient.java
```diff
@@ -122,26 +118,10 @@ public CertificateResponse SigningCert(CertificateRequest cr, String bearerToken
...
-      PemReader.Section section = pemReader.readNextSection();
-      if (section == null) {
-        break;
-      }
-
-      byte[] certBytes = section.getBase64DecodedBytes();
-      certList.add((X509Certificate) cf.generateCertificate(new ByteArrayInputStream(certBytes)));
+    try (InputStream content = resp.getContent()) {
+      return SigningCertificate.newSigningCertificate(
+          CharStreams.toString(new InputStreamReader(content)), sctHeader);
```
vlsi, line 124:
```suggestion
          CharStreams.toString(new InputStreamReader(content, resp.getContentCharset())), sctHeader);
```

src/main/java/dev/sigstore/fulcio/client/SigningCertificate.java
```diff
@@ -0,0 +1,117 @@
...
+    }
+
+    return cf.generateCertPath(certList);
+  }
+
+  @VisibleForTesting
+  static SignedCertificateTimestamp decodeSCT(String sctHeader) throws SerializationException {
+    byte[] sct = Base64.getDecoder().decode(sctHeader);
+    Gson gson = new GsonSupplier().get();
+    return gson.fromJson(new InputStreamReader(new ByteArrayInputStream(sct)), SctJson.class)
```
vlsi, line 69: I am not sure what SCT is, however, it looks like something should declare the encoding that is used to store SCT

```suggestion
    return gson.fromJson(new InputStreamReader(new ByteArrayInputStream(sct), StandardCharsets.UTF_8), SctJson.class)
```

loosebazooka: SCT are "signed certificate timestamps", they are authenticated metadata used to verify that a certifciate was logged with a certificate transparency log. The incoming SCT format is kind of fulcio specific, but it's definitely utf-8.

loosebazooka: I do think this will be changed to an embedded sct format in https://github.com/sigstore/fulcio/pull/507, but this is currently how in prod fulcio works

src/main/java/dev/sigstore/fulcio/client/FulcioValidator.java
```diff
@@ -0,0 +1,126 @@
...
+import org.conscrypt.ct.VerifiedSCT;
+
+public class FulcioValidator {
+  private final CTLogInfo ctLogInfo;
+  private final TrustAnchor fulcioRoot;
+
+  public static FulcioValidator newFulcioValidator(byte[] ctfePublicKey, byte[] fulcioRoot)
+      throws InvalidKeySpecException, NoSuchAlgorithmException, CertificateException, IOException,
+          InvalidAlgorithmParameterException {
+    KeyFactory keyFactory = KeyFactory.getInstance("EC");
```
haydentherapper, line 44: This actually wouldn't work for staging - Staging is using an RSA key. I don't think we should require that the CT log key type is an ECDSA key. Can this be configurable, or better yet, is there a way to easily try different key types? It will only be either an RSA, ECDSA, or EDDSA/ED25519 key.

See https://github.com/sigstore/cosign/pull/1806

src/main/java/dev/sigstore/fulcio/client/FulcioValidator.java
```diff
@@ -0,0 +1,126 @@
...
+    PemReader pemReader =
+        new PemReader(
+            new InputStreamReader(new ByteArrayInputStream(ctfePublicKey), StandardCharsets.UTF_8));
+    PemReader.Section section = pemReader.readNextSection();
+    if (pemReader.readNextSection() != null) {
+      throw new InvalidKeySpecException(
+          "ctfe public key must be only a single PEM encoded public key");
+    }
+    EncodedKeySpec publicKeySpec = new X509EncodedKeySpec(section.getBase64DecodedBytes());
+    PublicKey ctfePublicKeyObj = keyFactory.generatePublic(publicKeySpec);
```
haydentherapper, line 54: Related to the above, does Java handle multiple PEM encodings, both PKCS1 (BEGIN RSA PUBLIC KEY) or PKIX (BEGIN PUBLIC KEY)?

src/main/java/dev/sigstore/fulcio/client/FulcioValidator.java
```diff
@@ -0,0 +1,126 @@
...
+
+    CertificateFactory certificateFactory = CertificateFactory.getInstance("X.509");
+    X509Certificate fulcioRootObj =
+        (X509Certificate)
+            certificateFactory.generateCertificate(new ByteArrayInputStream(fulcioRoot));
+
+    TrustAnchor fulcioRootTrustAnchor = new TrustAnchor(fulcioRootObj, null);
+    // this should throw the InvalidAlgorithmException a bit earlier
+    new PKIXParameters(Collections.singleton(fulcioRootTrustAnchor));
```
haydentherapper, line 63: Does this need to be set to a variable?

src/main/java/dev/sigstore/fulcio/client/FulcioValidator.java
```diff
@@ -0,0 +1,126 @@
...
+
+    // TODO: this ctloginfo probably needs to be appropriately filled out
+    CTLogInfo ctLogInfo =
+        new CTLogInfo(ctfePublicKeyObj, "fulcio ct log", "intentionally-nonsense");
```
haydentherapper, line 67: random strings?

src/main/java/dev/sigstore/fulcio/client/FulcioValidator.java
```diff
@@ -0,0 +1,126 @@
...
+    return new FulcioValidator(ctLogInfo, fulcioRootTrustAnchor);
+  }
+
+  private FulcioValidator(CTLogInfo ctLogInfo, TrustAnchor fulcioRoot) {
+    this.ctLogInfo = ctLogInfo;
+    this.fulcioRoot = fulcioRoot;
```
haydentherapper, line 73: Does this handle intermediate certificates?

src/main/java/dev/sigstore/fulcio/client/FulcioValidator.java
```diff
@@ -0,0 +1,126 @@
...
+      throws CertificateEncodingException, FulcioValidationException {
+    if (!(sc.getLeafCertificate() instanceof X509Certificate)) {
+      throw new RuntimeException(
+          "Encountered non X509 Certificate when validating SCT. Leaf certificate is "
+              + sc.getLeafCertificate().getClass());
+    }
+    CertificateEntry ce =
+        CertificateEntry.createForX509Certificate((X509Certificate) sc.getLeafCertificate());
+
+    // a result is returned here, but I don't know what to do with it yet
```
haydentherapper, line 86: What's the options for statuses? I assume valid or not? If not valid, return an error?

src/main/java/dev/sigstore/fulcio/client/FulcioValidator.java
```diff
@@ -0,0 +1,126 @@
...
+    } catch (InvalidAlgorithmParameterException e) {
+      throw new RuntimeException(
+          "Can't create PKIX parameters for fulcioRoot. This should have been checked when generating a validator instance",
+          e);
+    }
+    pkixParams.setRevocationEnabled(false);
+
+    // these certs are only valid for 15 minutes, so find a time in the validity period
+    Date dateInValidityPeriod =
+        new Date(((X509Certificate) sc.getLeafCertificate()).getNotBefore().getTime() + 1000);
```
haydentherapper, line 116: Why adding time (+1000)?

src/main/java/dev/sigstore/fulcio/client/FulcioValidator.java
```diff
@@ -0,0 +1,126 @@
...
+    } catch (NoSuchAlgorithmException e) {
+      //
+      throw new RuntimeException(
+          "No PKIX CertPathValidator, we probably shouldn't be here, but this seems to be a system library error not a program control flow issue",
+          e);
+    }
+
+    PKIXParameters pkixParams;
+    try {
+      pkixParams = new PKIXParameters(Collections.singleton(fulcioRoot));
```
haydentherapper, line 106: So if I'm reading https://stackoverflow.com/questions/60633743/should-intermediate-certificates-be-added-as-trustanchors-when-validating-certif correctly, it seems like intermediates are handled automatically?

src/main/java/dev/sigstore/fulcio/client/SigningCertificate.java
```diff
@@ -0,0 +1,120 @@
...
+      }
+
+      DigitallySigned digiSig = DigitallySigned.decode(signature);
+      return new SignedCertificateTimestamp(
+          SignedCertificateTimestamp.Version.V1,
+          id,
+          timestamp,
+          extensions,
+          digiSig,
+          SignedCertificateTimestamp.Origin.OCSP_RESPONSE);
```
haydentherapper, line 100: This is for the detached SCT as a header, correct? Will there be a later PR for handling embedded?

src/main/java/dev/sigstore/fulcio/client/SigningCertificate.java
```diff
@@ -0,0 +1,120 @@
...
+    this.certPath = certPath;
+    this.sct = sct;
+  }
+
+  public CertPath getCertPath() {
+    return certPath;
+  }
+
+  public Certificate getLeafCertificate() {
+    return certPath.getCertificates().get(0);
```
haydentherapper, line 114: Should this check that SigningCertificate was initialized with at least one certificate in the cert path?

src/main/java/dev/sigstore/fulcio/client/FulcioValidator.java
```diff
@@ -0,0 +1,126 @@
...
+    this.ctLogInfo = ctLogInfo;
+    this.fulcioRoot = fulcioRoot;
+  }
+
+  public void validateSct(SigningCertificate sc)
+      throws CertificateEncodingException, FulcioValidationException {
+    if (!(sc.getLeafCertificate() instanceof X509Certificate)) {
+      throw new RuntimeException(
+          "Encountered non X509 Certificate when validating SCT. Leaf certificate is "
+              + sc.getLeafCertificate().getClass());
```
vlsi, line 81: It would probably make sense to expand what SCT is since the ones who will run into the error (e.g. end-users) would likely be unfamiliar with SCTs

src/main/java/dev/sigstore/fulcio/client/FulcioValidator.java
loosebazooka: interesting, I'll add a todo and file a bug.

src/main/java/dev/sigstore/fulcio/client/FulcioValidator.java
loosebazooka: Yeah this PEM reader doesn't care, it seems to match against `-----BEGIN ([A-Z ]+)-----`. The keyFactory will probably fail if no ecdsa though until the earlier comment is fixed.

src/main/java/dev/sigstore/fulcio/client/FulcioValidator.java
loosebazooka: that was to trigger an exception earlier, so the program fails in a more meaningful way when things are being setup.
see: https://github.com/sigstore/sigstore-java/pull/1/files/d3f3635533411d76dc19c9d8fdb7c2a08054d5c9#diff-c9c3437da746d8734c6a12c05a0c712053a0c672c309cef8243b6cf83128c6abR109

src/main/java/dev/sigstore/fulcio/client/FulcioValidator.java
loosebazooka: There's a todo sitting there, because I think this needs to be configurable based on the CTLog that fulcio is configured to be using (staging, prod, user-custom)

src/main/java/dev/sigstore/fulcio/client/FulcioValidator.java
loosebazooka: yeah I have some code sitting on my machine using the new fulcio intermediates that seems to work. But I think that just verifies the chain back up to the single fulcio root. Not the intermediates, I'm not exactly sure what needs to be done here though?

src/main/java/dev/sigstore/fulcio/client/FulcioValidator.java
loosebazooka: I expect this to be a programming or setup error that most end users wouldn't encounter.

src/main/java/dev/sigstore/fulcio/client/FulcioValidator.java
loosebazooka: If not valid, it throws an exception, but I'm mucking around with internal APIs of this library. I'm trying to see if they will offer an API for this kind of stuff. Other options are:
https://github.com/google/conscrypt/blob/master/common/src/main/java/org/conscrypt/ct/VerifiedSCT.java#L25

src/main/java/dev/sigstore/fulcio/client/FulcioValidator.java
loosebazooka: yeah, when testing with intermediates, this appears to hold true.

src/main/java/dev/sigstore/fulcio/client/FulcioValidator.java
loosebazooka: it was arbitrary amount of time. Not sure I have an actual reason, lemme test without.

src/main/java/dev/sigstore/fulcio/client/SigningCertificate.java
loosebazooka: yup

src/main/java/dev/sigstore/fulcio/client/SigningCertificate.java
loosebazooka: the initializer has already verified this: https://github.com/sigstore/sigstore-java/pull/1/files/d3f3635533411d76dc19c9d8fdb7c2a08054d5c9#diff-98c6d9ebfb8758548d445c2b02eadea2f97338e0104dcec6670245c70255b395R60

src/main/java/dev/sigstore/fulcio/client/FulcioValidator.java
loosebazooka: kk 1000 is unnecessary

src/main/java/dev/sigstore/fulcio/client/SigningCertificate.java
loosebazooka: So this would throw an out of bounds runtime exception which would reveal a programming error.

src/main/java/dev/sigstore/fulcio/client/FulcioClient.java
```diff
@@ -122,26 +118,11 @@ public CertificateResponse SigningCert(CertificateRequest cr, String bearerToken
     if (sctHeader == null) {
```
haydentherapper, line 118: Fulcio doesn't have to send the SCT if the log isn't configured to send it - https://github.com/sigstore/fulcio/blob/main/pkg/api/legacy_server.go#L111-L119

src/main/java/dev/sigstore/fulcio/client/FulcioValidator.java
haydentherapper: For the third parameter, we could provide a valid URL to the CT log, but I'm hesitant to do so - It's a sharded log, so at some point we'd have to upgrade the java fulcio client to point to the new log. The URL is probably used to confirm the presence of an entry for non-SCT verification. Maybe change `intentionally-nonsense` to `unused-url`?

src/main/java/dev/sigstore/fulcio/client/SigningCertificate.java
```diff
@@ -0,0 +1,120 @@
...
+    SignedCertificateTimestamp sct = decodeSCT(sctHeader);
+    return new SigningCertificate(certPath, sct);
+  }
+
+  @VisibleForTesting
+  static CertPath decodeCerts(String content) throws CertificateException, IOException {
+    PemReader pemReader = new PemReader(new StringReader(content));
+    CertificateFactory cf = CertificateFactory.getInstance("X.509");
+    ArrayList<X509Certificate> certList = new ArrayList<>();
+    while (true) {
```
haydentherapper, line 50: Might be worth considering a maximum PEM length - Otherwise it's a DOS vector, if I provide you a really long certificate chain. I think it's reasonable to fail after, say, 10 iterations?

src/main/java/dev/sigstore/fulcio/client/SigningCertificate.java
```diff
@@ -0,0 +1,120 @@
...
+  static SigningCertificate newSigningCertificate(String certs, String sctHeader)
+      throws CertificateException, IOException, SerializationException {
+    CertPath certPath = decodeCerts(certs);
+    SignedCertificateTimestamp sct = decodeSCT(sctHeader);
+    return new SigningCertificate(certPath, sct);
+  }
+
+  @VisibleForTesting
+  static CertPath decodeCerts(String content) throws CertificateException, IOException {
+    PemReader pemReader = new PemReader(new StringReader(content));
```
haydentherapper, line 47: Does this fail if a DER encoded certificate is provided? (I think it should)

loosebazooka: actually, the request sets the header to only accept application/pem-certificate-chain, I mean it still fails, so I could write a test to define this behavior, but it would be weird if we got to that point.

loosebazooka: hrmm, maybe that accept type doesn't actually mean anything... I've added the test anyway

src/main/java/dev/sigstore/fulcio/client/FulcioClient.java
loosebazooka: even if you can technically configure the fulcio server in no ct log mode `--ct-log-url=`

src/main/java/dev/sigstore/fulcio/client/FulcioValidator.java
loosebazooka: The new code will handle intermediates.

src/main/java/dev/sigstore/fulcio/client/FulcioValidator.java
loosebazooka: filed at #4

src/main/java/dev/sigstore/fulcio/client/FulcioClient.java
haydentherapper: It would be valid, if fulcio is self hosted for example. For the public insurance, there should always be an sct

src/main/java/dev/sigstore/fulcio/client/SigningCertificate.java
haydentherapper: Nvm, I was thinking this is running server side for some reason. There’s no risk here

src/main/java/dev/sigstore/fulcio/client/FulcioClient.java
loosebazooka: kk, I'll make it a config option on the client

loosebazooka: done

src/main/java/dev/sigstore/fulcio/client/FulcioValidationException.java
```diff
@@ -15,25 +15,16 @@
...
-public class CertificateResponse {
-  private final CertPath certPath;
-
-  // TODO: This could be saved potentially as a more concrete type
-  @Nullable private final byte[] sct;
-
-  public CertificateResponse(CertPath certPath, @Nullable byte[] sct) {
-    this.certPath = certPath;
-    this.sct = sct;
+public class FulcioValidationException extends Exception {
```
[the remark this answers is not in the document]
patflynn: Should this extend RuntimeException?

src/main/java/dev/sigstore/fulcio/client/SigningCertificate.java
```diff
@@ -0,0 +1,125 @@
...
+      PemReader.Section section = pemReader.readNextSection();
+      if (section == null) {
+        break;
+      }
+
+      byte[] certBytes = section.getBase64DecodedBytes();
+      certList.add((X509Certificate) cf.generateCertificate(new ByteArrayInputStream(certBytes)));
+    }
+    if (certList.isEmpty()) {
+      throw new IOException("no valid PEM certificates were found in response from Fulcio");
```
patflynn, line 65: Is this maybe like an IllegalArgument or IllegalFormat exception? I have no idea, but IO feels a little weird since the IO has happened but the contents are not as expected.

src/test/java/dev/sigstore/fulcio/client/FulcioClientTest.java
```diff
@@ -65,11 +67,60 @@ public void testSigningCert() throws Exception {
...
+  }
+
+  @Test
+  public void testSigningCert_NoSct() throws Exception {
+    try (FakeOIDCServer oidcServer = FakeOIDCServer.startNewServer()) {
+      File fulcioConfig = testRoot.newFile("fulcio-config.json");
+      Files.write(oidcServer.getFulcioConfig().getBytes(StandardCharsets.UTF_8), fulcioConfig);
+      // start fulcio client with config from oidc server
+      FulcioWrapper fulcioServer = null;
+      System.out.println("Start new server");
```
patflynn, line 91: remind me, is system.out.println typical from a test or is a logger used? I have no idea, I'm just curious and my google-foo is letting me down.

patflynn (approved):
Lots of mysterious stuff in here, but looks like you know what you're doing! 

My meta comment is that this would be easier to review broken up into neater PRs. For instance the refactorings could have been separated out. No biggie though.

src/main/java/dev/sigstore/fulcio/client/FulcioClient.java
```diff
@@ -40,22 +37,26 @@ public class Client {
...
-  private Client(HttpTransport httpTransport, URI serverUrl, String userAgent) {
+  private FulcioClient(
+      HttpTransport httpTransport, URI serverUrl, String userAgent, boolean requireSct) {
     this.httpTransport = httpTransport;
     this.serverUrl = serverUrl;
     this.userAgent = userAgent;
+    this.requireSct = requireSct;
   }
 
   public static class Builder {
```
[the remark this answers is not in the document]
loosebazooka: Yeah, so that makes sense, and I'll add those in before I do any new work

src/main/java/dev/sigstore/fulcio/client/FulcioValidator.java
```diff
@@ -0,0 +1,149 @@
...
+import org.conscrypt.ct.SignedCertificateTimestamp;
+import org.conscrypt.ct.VerifiedSCT;
+
+public class FulcioValidator {
+  @Nullable private final CTLogInfo ctLogInfo;
+  private final TrustAnchor fulcioRoot;
+
+  public static FulcioValidator newFulcioValidator(
+      byte[] fulcioRoot, @Nullable byte[] ctfePublicKey)
+      throws InvalidKeySpecException, NoSuchAlgorithmException, CertificateException, IOException,
```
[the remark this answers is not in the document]
loosebazooka: it can be null because scts aren't required to be returned from fulcio. they *should* be, but people could be running private instances and maybe don't care about internal certificate transparency (but maybe they should?)?

src/main/java/dev/sigstore/fulcio/client/SigningCertificate.java
loosebazooka: yeah it's def not an IOException, lemme fix that.

src/test/java/dev/sigstore/fulcio/client/FulcioClientTest.java
loosebazooka: oh yeah this is from debugging, I couldn't for the life of me over the last two days get consecutive fulcio instances up while running these tests.

src/main/java/dev/sigstore/fulcio/client/FulcioValidator.java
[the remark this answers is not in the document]
haydentherapper: Can this be enforced via a policy/config?

haydentherapper (dismissed):

src/main/java/dev/sigstore/fulcio/client/FulcioValidationException.java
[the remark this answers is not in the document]
patflynn: OK. Makes sense. We should chat about our approach here some time with Jason. I'll typically avoid throwing any checked exceptions unless it's type or api enables the client in some way.

src/main/java/dev/sigstore/fulcio/client/FulcioValidator.java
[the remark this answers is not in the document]
patflynn: But validate will always fail as a result won't it?

patflynn (approved):

PR #5, opened 2022-05-08
Add OIDC client

loosebazooka:
This is a convenience wrapper around the google oauth client, with
specific defaults for sigstore AND automatic endpoint detection
from the well known issuer open id configuration url (which
strangely is not part of the oauth client).

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -20,17 +20,24 @@ tasks.compileTestJava {
 dependencies {
     implementation(platform("com.google.cloud:libraries-bom:24.3.0"))
     implementation("com.google.http-client:google-http-client-apache-v2")
-    implementation("com.google.api-client:google-api-client-gson:1.31.5")
+    implementation("com.google.http-client:google-http-client-gson")
 
     implementation("com.google.code.gson:gson:2.8.9")
     implementation("org.conscrypt:conscrypt-openjdk-uber:2.5.2") {
         because("contains library code for all platforms")
     }
 
+    implementation(platform("com.google.oauth-client:google-oauth-client-bom:1.33.3"))
+    implementation("com.google.oauth-client:google-oauth-client")
+    implementation("com.google.oauth-client:google-oauth-client-jetty")
+    implementation("com.google.oauth-client:google-oauth-client-java6")
+
     testImplementation("junit:junit:4.12")
     testImplementation("com.nimbusds:oauth2-oidc-sdk:6.21.2")
     testImplementation("com.nimbusds:nimbus-jose-jwt:9.18")
+    testImplementation("no.nav.security:mock-oauth2-server:0.4.4")
     testImplementation("org.bouncycastle:bcutil-jdk15on:1.70")
+    testImplementation("net.sourceforge.htmlunit:htmlunit:2.61.0")
 }
 
 spotless {
--- a/src/main/java/dev/sigstore/http/HttpProvider.java
+++ b/src/main/java/dev/sigstore/http/HttpProvider.java
@@ -1,0 +1,88 @@
+/*
+ * Copyright 2022 The Sigstore Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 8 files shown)


build.gradle.kts
```diff
@@ -20,14 +20,19 @@ tasks.compileTestJava {
 dependencies {
     implementation(platform("com.google.cloud:libraries-bom:24.3.0"))
     implementation("com.google.http-client:google-http-client-apache-v2")
-    implementation("com.google.api-client:google-api-client-gson:1.31.5")
+    implementation("com.google.http-client:google-http-client-gson")
```
patflynn, line 23: (more noob questions)
should we not pin versions?

src/main/java/dev/sigstore/http/HttpProvider.java
```diff
@@ -0,0 +1,81 @@
...
+ */
+package dev.sigstore.http;
+
+import com.google.api.client.http.HttpTransport;
+import com.google.api.client.http.apache.v2.ApacheHttpTransport;
+import java.util.concurrent.TimeUnit;
+import org.apache.http.conn.ssl.NoopHostnameVerifier;
+import org.apache.http.impl.client.HttpClientBuilder;
+
+public class HttpProvider {
```
patflynn, line 24: I'm not sure but I thought the Http stuff was an implementation detail in the chat we had. Does it need to be public?

src/main/java/dev/sigstore/oidc/client/OidcClient.java
```diff
@@ -0,0 +1,252 @@
...
+      return new OidcClient(hp, issuer, clientId, bh);
+    }
+  }
+
+  /**
+   * Get an id token from the oidc provider
+   *
+   * @param expectedEmailAddress optional parameter to verify the subject email address does indeed
+   *     match the value in the id token
+   * @return
```
patflynn, line 111: missing javadoc

src/main/java/dev/sigstore/oidc/client/OidcClient.java
```diff
@@ -0,0 +1,252 @@
...
+      if (!idTokenVerifier.verify(parsedIdToken)) {
+        throw new OidcException("id token could not be verified");
+      }
+    } catch (IOException e) {
+      // TODO: maybe a more descriptive exception message
+      throw new OidcException("ioexception during oidc handshake", e);
+    }
+
+    String emailFromIDToken = (String) parsedIdToken.getPayload().get("email");
+    Boolean emailVerified = (Boolean) parsedIdToken.getPayload().get("email_verified");
```
patflynn, line 166: I'm out of date probably, but why does this have to be a non primitive type?

src/main/java/dev/sigstore/oidc/client/OidcException.java
```diff
@@ -0,0 +1,26 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package dev.sigstore.oidc.client;
+
+public class OidcException extends Exception {
```
patflynn, line 18: I guess we create these so that we can specialize the exceptions in the future?   I remember back in the day, the cool way to throw exceptions was to use a subtype of RuntimeException unless the caller was likely to be able to handle the issue.

build.gradle.kts
loosebazooka: versions are in the `bom`

patflynn (approved):

src/main/java/dev/sigstore/http/HttpProvider.java
loosebazooka: This is supposed to be the wrapper around the implementation. So we can swap out the underlying implementation later if we want.

loosebazooka: I kind of changed this around in #8

loosebazooka: Ideally we do something that lets the user inject their http client of choice. But that's definitely for another day.

patflynn (approved):

src/main/java/dev/sigstore/oidc/client/OidcException.java
loosebazooka: Yeah I do. lulz

loosebazooka: I meant runtime environment/programming error. Not the jdk.

PR #6, opened 2022-05-10
Use Java 11 in tests

loosebazooka:
Also fixes fake CT log with response that is compatible
with what fulcio is expecting (otherwise test fails).


loosebazooka:
I need to possibly properly kill fulcio in #1 if I run multiple instances of fulcio wrapper.

patflynn (approved):

PR #20, opened 2022-05-18
upgrade to newer google java format

loosebazooka:
since we moved to java11 we are no longer stuck on 1.6


loosebazooka:
ugh: https://github.com/google/google-java-format/issues/779

patflynn (dismissed):

patflynn:
I suspect you have some kind of java version mismatch

loosebazooka:
I think some features that new java format wants to use and not exported in 17....

loosebazooka:
lets try with 1.13.0 instead of 1.15.0

loosebazooka:
uh, I just removed the conflicting component with okhttps mockwebserver, we should be good. This was built for testing anyway. So we come out on top!

loosebazooka:
okay this no longer has anything to do with upgrading google java format, it's just replacing the jre test webserver with mockhttpserver

patflynn (approved):

PR #21, opened 2022-05-18
Fail if we can't parse a key and bc isn't present

loosebazooka:

Auto-adding bc scared me a bit, wdyt about this?

### DIFF
--- a/src/main/java/dev/sigstore/encryption/Keys.java
+++ b/src/main/java/dev/sigstore/encryption/Keys.java
@@ -15,7 +15,6 @@
  */
 package dev.sigstore.encryption;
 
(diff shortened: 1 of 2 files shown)


loosebazooka:
I don't have a great answer here. But I err on the side of caution?

patflynn:
Before we merge this I'd like to see some general recommendation that we shouldn't do it.  I couldn't find anything online, but maybe you can check with the java illuminati at Google?

loosebazooka:
oh this PR is mostly a question, so yeah lemme change it to a draft

loosebazooka:
or alternatively can we just pull it in if we encounter an eddsa key?

PR #36, opened 2022-06-02
Support Fulcio certs with embedded SCTs

loosebazooka:
Also update verifier to be able to verify them


src/main/java/dev/sigstore/fulcio/client/FulcioVerifier.java
```diff
@@ -59,42 +65,76 @@ public static FulcioVerifier newFulcioVerifier(byte[] fulcioRoot, byte @Nullable
...
     this.fulcioRoot = fulcioRoot;
+    this.ctLogInfo = ctLogInfo;
+    // we are exploiting some internal objects here. A CTVerifier needs to lookup which logs
+    // apply to which scts. Since we only have a single sct (the fulcio sct), we initialize
+    // the "CTVerifier" with an anonymous implementation of CTLogStore with a single matcher
+    // for our ctLogInfo (i.e. if the sct is signed with the ctfe public key, we can verify it).
+    this.ctVerifier =
+        (ctLogInfo == null)
+            ? null
+            : new CTVerifier(logId -> Arrays.equals(logId, ctLogInfo.getID()) ? ctLogInfo : null);
```
patflynn, line 77: Maybe I'm reading this wrong, but are you setting the CTVerifier store to null if the logs aren't equal?  Will that cause failures later on?

src/main/java/dev/sigstore/fulcio/client/FulcioVerifier.java
```diff
@@ -59,42 +65,76 @@ public static FulcioVerifier newFulcioVerifier(byte[] fulcioRoot, byte @Nullable
...
+      CTVerificationResult result;
+      try {
+        result = ctVerifier.verifySignedCertificateTimestamps(certs, null, null);
+      } catch (CertificateEncodingException cee) {
+        throw new FulcioVerificationException(
+            "Certificates could not be parsed during sct verification");
+      }
+      int valid = result.getValidSCTs().size();
+      int invalid = result.getInvalidSCTs().size();
+      if (valid != 1 || invalid != 0) {
```
haydentherapper, line 116: i wouldn't hardcode valid to 1. at some point, we may start including multiple SCTs from multiple logs in certificates. you just need to iterate through all SCTs and verify each

src/main/java/dev/sigstore/fulcio/client/SigningCertificate.java
```diff
@@ -79,6 +81,10 @@ static SignedCertificateTimestamp decodeSCT(String sctHeader) throws Serializati
         .toSct();
   }
 
+  public boolean hasEmbeddedSct() {
```
patflynn, line 84: some javadoc about embedded vs detached SCT would help here.

src/main/java/dev/sigstore/fulcio/client/FulcioVerifier.java
loosebazooka: Makes sense.

But, on the topic of multiple logs -- where would these logs be? Will the logs public key be distributed via tuf root? I'm not sure how this works.

src/main/java/dev/sigstore/fulcio/client/FulcioVerifier.java
loosebazooka: That's a shortcut to implementing [CTLogStore](https://github.com/google/conscrypt/blob/ca6485b689219323e91d3bbe0410715b5b2d687f/common/src/main/java/org/conscrypt/ct/CTLogStore.java)

It might be a little confusing, I'll expand out the lambda.

src/main/java/dev/sigstore/fulcio/client/FulcioVerifier.java
haydentherapper: Unclear at this point, maybe delegated keys

src/main/java/dev/sigstore/fulcio/client/FulcioClient.java
```diff
@@ -111,14 +111,16 @@ public SigningCertificate SigningCert(CertificateRequest cr)
...
-
     try (InputStream content = resp.getContent()) {
-      return SigningCertificate.newSigningCertificate(
-          CharStreams.toString(new InputStreamReader(content, resp.getContentCharset())),
-          sctHeader);
+      var signingCert =
+          SigningCertificate.newSigningCertificate(
+              CharStreams.toString(new InputStreamReader(content, resp.getContentCharset())),
+              sctHeader);
+      if (signingCert.getDetachedSct().isEmpty() && !signingCert.hasEmbeddedSct() && requireSct) {
```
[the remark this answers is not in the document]
haydentherapper: Testing or a private instance that doesn’t want auditability

src/main/java/dev/sigstore/fulcio/client/SigningCertificate.java
loosebazooka: added to the doc on the class.

patflynn (approved):

PR #37, opened 2022-06-03
add get and search for rekor entries.

patflynn:
WIP adding basic read operations for Rekor. For now this is a raw translation of the rest API, we can chat a bit about use-cases and refactor, just wanted to get this out there to make sure it's not way off base.  

Testing: ideally tests would be against a clean rekor environment, I have placeholders for now just to make sure the API was working as expected. I'll update the tests to create some data in staging following the pattern already established.

### DIFF
--- a/src/main/java/dev/sigstore/json/GsonSupplier.java
+++ b/src/main/java/dev/sigstore/json/GsonSupplier.java
@@ -17,6 +17,8 @@ package dev.sigstore.json;
 
 import com.google.gson.Gson;
 import com.google.gson.GsonBuilder;
+import dev.sigstore.rekor.client.GsonAdaptersHashedRekordWrapper;
+import dev.sigstore.rekor.client.GsonAdaptersRekorEntry;
 import java.util.function.Supplier;
 
 /**
@@ -30,6 +32,8 @@ public class GsonSupplier implements Supplier<Gson> {
   public Gson get() {
     return new GsonBuilder()
         .registerTypeAdapter(byte[].class, new GsonByteArrayAdapter())
+        .registerTypeAdapterFactory(new GsonAdaptersRekorEntry())
+        .registerTypeAdapterFactory(new GsonAdaptersHashedRekordWrapper())
         .disableHtmlEscaping()
         .create();
   }
--- a/src/main/java/dev/sigstore/rekor/client/RekorClient.java
+++ b/src/main/java/dev/sigstore/rekor/client/RekorClient.java
@@ -18,11 +18,13 @@ package dev.sigstore.rekor.client;
 import com.google.api.client.http.*;
 import dev.sigstore.http.HttpProvider;
 import dev.sigstore.json.GsonSupplier;
+
 import java.io.IOException;
 import java.net.URI;
 import java.util.Arrays;
 import java.util.HashMap;
 import java.util.List;
+import java.util.Optional;
 
 /** A client to communicate with a rekor service instance. */
 public class RekorClient {
@@ -98,7 +100,7 @@ public class RekorClient {
     return RekorResponse.newRekorResponse(rekorEntryUri, entry);
   }
 
-  public RekorEntry getEntry(String UUID) throws IOException {
+  public Optional<RekorEntry> getEntry(String UUID) throws IOException {
     URI getEntryURI = serverUrl.resolve(REKOR_ENTRIES_PATH + "/" + UUID);
     HttpRequest req =
         httpProvider
@@ -110,10 +112,10 @@ public class RekorClient {
     try {
       response = req.execute();
     } catch (HttpResponseException e) {
-      if (e.getStatusCode() == 404) return null;
+      if (e.getStatusCode() == 404) return Optional.ofNullable(null);
       throw e;
     }
-    return RekorResponse.newRekorResponse(getEntryURI, response.parseAsString()).getEntry();
+    return Optional.of(RekorResponse.newRekorResponse(getEntryURI, response.parseAsString()).getEntry());
   }
 
   /**
--- a/src/main/java/dev/sigstore/rekor/client/RekorClient.java
+++ b/src/main/java/dev/sigstore/rekor/client/RekorClient.java
@@ -18,7 +18,6 @@ package dev.sigstore.rekor.client;
 import com.google.api.client.http.*;
 import dev.sigstore.http.HttpProvider;
 import dev.sigstore.json.GsonSupplier;
-
 import java.io.IOException;
 import java.net.URI;
 import java.util.Arrays;
@@ -115,7 +114,8 @@ public class RekorClient {
       if (e.getStatusCode() == 404) return Optional.ofNullable(null);
       throw e;
     }
-    return Optional.of(RekorResponse.newRekorResponse(getEntryURI, response.parseAsString()).getEntry());
+    return Optional.of(
+        RekorResponse.newRekorResponse(getEntryURI, response.parseAsString()).getEntry());
   }
 
   /**
--- a/src/main/java/dev/sigstore/rekor/client/RekorClient.java
+++ b/src/main/java/dev/sigstore/rekor/client/RekorClient.java
@@ -18,7 +18,6 @@ package dev.sigstore.rekor.client;
 import com.google.api.client.http.*;
 import dev.sigstore.http.HttpProvider;
 import dev.sigstore.json.GsonSupplier;
-
 import java.io.IOException;
 import java.net.URI;
 import java.util.Arrays;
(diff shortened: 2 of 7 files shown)


src/main/java/dev/sigstore/rekor/client/RekorClient.java
```diff
@@ -15,18 +15,21 @@
...
 import java.net.URI;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.List;
 
 /** A client to communicate with a rekor service instance. */
 public class RekorClient {
   public static final String PUBLIC_REKOR_SERVER = "https://rekor.sigstore.dev";
   public static final String REKOR_ENTRIES_PATH = "/api/v1/log/entries";
+  public static final String REKOR_ENTRIES_GET_PATH = "/api/v1/log/entries/";
```
loosebazooka, line 31: presumably you already tried this, but there's no way to dedup this?

src/main/java/dev/sigstore/rekor/client/RekorClient.java
```diff
@@ -88,16 +91,69 @@ public RekorResponse putEntry(HashedRekordRequest hashedRekordRequest) throws IO
...
+        httpProvider
+            .getHttpTransport()
+            .createRequestFactory()
+            .buildGetRequest(new GenericUrl(getEntryURI));
+    req.getHeaders().set("Accept", "application/json");
+    HttpResponse response;
+    try {
+      response = req.execute();
+    } catch (HttpResponseException e) {
+      if (e.getStatusCode() == 404) return null;
```
loosebazooka, line 114: should this still be an error in the `get` workflow.

src/main/java/dev/sigstore/rekor/client/RekorClient.java
```diff
@@ -88,16 +91,69 @@ public RekorResponse putEntry(HashedRekordRequest hashedRekordRequest) throws IO
...
+      throw e;
+    }
+    return RekorResponse.newRekorResponse(getEntryURI, response.parseAsString()).getEntry();
```
loosebazooka, line 117: Does this response not include an inclusion proof?

src/main/java/dev/sigstore/rekor/client/RekorClient.java
```diff
@@ -88,16 +91,69 @@ public RekorResponse putEntry(HashedRekordRequest hashedRekordRequest) throws IO
...
+
+  /**
+   * Returns a list of UUIDs for matching entries for the given search parameters.
+   *
+   * @param email the OIDC email subject
+   * @param hash sha256 hash of the artifact
+   * @param publicKeyFormat format of public key (one of 'pgp','x509','minisign', 'ssh', 'tuf')
+   * @param publicKeyContent public key base64 encoded content
+   */
+  public List<String> searchEntry(
```
loosebazooka, line 128: Is the workflow supposed to be?
1. search of UUID
2. get entry by UUID

src/main/java/dev/sigstore/rekor/client/RekorClient.java
```diff
@@ -88,16 +91,69 @@ public RekorResponse putEntry(HashedRekordRequest hashedRekordRequest) throws IO
...
 
     URI rekorEntryUri = serverUrl.resolve(resp.getHeaders().getLocation());
     String entry = resp.parseAsString();
     return RekorResponse.newRekorResponse(rekorEntryUri, entry);
   }
 
-  /** Obtain an entry for an artifact from rekor. */
-  public void getEntry() {
-    throw new UnsupportedOperationException("I'm a worthless download function");
+  public RekorEntry getEntry(String UUID) throws IOException {
```
loosebazooka, line 102: if we can return null, this should be `Optional<RekorEntry>`

src/main/java/dev/sigstore/rekor/client/RekorClient.java
patflynn: That's what the web UI does, it seems like the only way to retrieve a full rekord without a uuid. That doesn't mean we can't present a nicer API to clients.

loosebazooka: Word. I didn't realize that's how it worked

src/main/java/dev/sigstore/rekor/client/RekorClient.java
patflynn: I believe it does, we just need to update our gson type adapter to deserialize it.

src/main/java/dev/sigstore/rekor/client/RekorClient.java
patflynn: ¯\_(ツ)_/¯. Optional or NFE, happy to do either.

src/main/java/dev/sigstore/rekor/client/RekorClient.java
patflynn: yeah I can just construct the path.

src/main/java/dev/sigstore/rekor/client/RekorEntry.java
```diff
@@ -15,13 +15,21 @@
...
+import dev.sigstore.rekor.Hashedrekord;
 import org.immutables.gson.Gson;
 import org.immutables.value.Value;
 
 /** A local representation of a rekor entry in the log. */
 @Gson.TypeAdapters
 @Value.Immutable
 public interface RekorEntry {
+
+  com.google.gson.Gson GSON = new GsonSupplier().get();
```
loosebazooka, line 31: `GSON` -> `gson`
I think you can clean up the fully qualified name and just import it?

loosebazooka: oh it's a const... nvm?

src/test/java/dev/sigstore/rekor/client/RekorClientTest.java
```diff
@@ -26,20 +30,137 @@
...
+    assertTrue(entry.isEmpty());
+  }
+
+  @NotNull
+  private HashedRekordRequest createdRekorRequest()
+      throws NoSuchAlgorithmException, InvalidKeyException, SignatureException,
+          OperatorCreationException, CertificateException, IOException, URISyntaxException {
     // the data we want to sign
-    var data = "some data";
+    var data = "some data " + System.currentTimeMillis();
```
loosebazooka, line 163: you can also use UUID.randomUUID() but I dunno if it matters.

src/main/java/dev/sigstore/json/GsonSupplier.java
```diff
@@ -30,6 +32,8 @@ public class GsonSupplier implements Supplier<Gson> {
   public Gson get() {
```
loosebazooka, line 32: maybe we should just singleton this. I thought we might be injecting this so I made it instantiable(?), but it's not clear we'll ever do that. But this is probably for another PR.

src/main/java/dev/sigstore/rekor/client/RekorClient.java
```diff
@@ -88,16 +91,70 @@ public RekorResponse putEntry(HashedRekordRequest hashedRekordRequest) throws IO
...
+        httpProvider
+            .getHttpTransport()
+            .createRequestFactory()
+            .buildGetRequest(new GenericUrl(getEntryURI));
+    req.getHeaders().set("Accept", "application/json");
+    HttpResponse response;
+    try {
+      response = req.execute();
+    } catch (HttpResponseException e) {
+      if (e.getStatusCode() == 404) return Optional.ofNullable(null);
```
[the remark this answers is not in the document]
patflynn: 404 is explicitly for resource not found. I believe it's the standard response for any restful services to indicate absence rather than any error. That being said I don't feel strongly either way, I'm leaning towards this approach based off exceptions typically being reserved for error conditions and not expected control flow. For instance that's why normal map behaviour is to return null for key look ups that aren't present.

src/main/java/dev/sigstore/rekor/client/RekorEntry.java
patflynn: ha, I hadn't even realized I'd introduced a field on an interface. Now I have to read-up on how that works... for now I'm going to in-line, it doesn't make sense as a public part of the interface.  Maybe we can think about making it a singleton later?

src/test/java/dev/sigstore/rekor/client/RekorClientTest.java
patflynn: done.

src/main/java/dev/sigstore/rekor/client/RekorClient.java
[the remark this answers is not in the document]
loosebazooka: Yeah you're right, I was thinking of 500 for some reason. nvm.

src/main/java/dev/sigstore/json/GsonSupplier.java
loosebazooka: #42

loosebazooka (approved):

PR #40, opened 2022-06-06
Add maven-publish plugin

loosebazooka:
Add's the maven publish plugin with only a mechanism for locally installing, this will need to be extended to publish to maven central (but it seems odd to publish this plugin without signing it with sigstore?) So lets see how that goes -- maybe we could do it with a non-gradle tool for now and just include it in the publication.

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -7,3 +7,8 @@
 ```
 ./gradlew spotlessApply
 ```
+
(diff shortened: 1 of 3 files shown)


patflynn (approved):
how will we sign our first central publish? :)

loosebazooka:
gonna have to cosign it

PR #43, opened 2022-06-09
Add fulcio grpc definition and generation

loosebazooka:
Add in proto def from fulcio and add code generation during build.


patflynn (approved):

patflynn:
Eventually WDYT about having the fulcio repo generate and publish the Java grpc artifact so that we can import it?

PR #46, opened 2022-06-10
E2e sign

loosebazooka:
@patflynn FYI

requires #53


loosebazooka:
needs #56 before CI will pass

src/main/java/dev/sigstore/KeylessSigner.java
```diff
@@ -0,0 +1,185 @@
...
+import java.net.URI;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.security.*;
+import java.security.cert.CertificateException;
+import java.security.spec.InvalidKeySpecException;
+import org.conscrypt.ct.SerializationException;
+
+public class KeylessSigner {
```
patflynn, line 40: javadoc with an example flow could be cool here.

src/main/java/dev/sigstore/KeylessSigner.java
```diff
@@ -0,0 +1,185 @@
...
+      Signer signer) {
+    this.fulcioClient = fulcioClient;
+    this.fulcioVerifier = fulcioVerifier;
+    this.rekorClient = rekorClient;
+    this.rekorVerifier = rekorVerifier;
+    this.oidcClient = oidcClient;
+    this.signer = signer;
+  }
+
+  public static Builder builderForProd()
```
patflynn, line 63: I wonder if prod is the right wording here. Presumably users will be configuring this for either their own private sigstore infra, or one of the public providers.  Perhaps something more specific like sigstorePublicGood?  

I'm also wondering about usescases for doing something with this builder prior to building? In the tests it just doesBuilderForProd().build(), but why not cut out the middleperson?

src/main/java/dev/sigstore/KeylessSigner.java
```diff
@@ -0,0 +1,185 @@
...
+      throws IOException, InvalidAlgorithmParameterException, CertificateException,
+          InvalidKeySpecException, NoSuchAlgorithmException {
+    var builder = new Builder();
+    var fulcioCert =
+        Resources.toByteArray(
+            Resources.getResource("dev/sigstore/tuf/production/fulcio_v1.crt.pem"));
```
patflynn, line 69: I still haven't got the TUF stuff straight in my head, so I'm curious how a key rotation would work here.

src/main/java/dev/sigstore/KeylessSigner.java
```diff
@@ -0,0 +1,185 @@
...
+        .fulcioClient(
+            FulcioClient.builder().build(),
+            FulcioVerifier.newFulcioVerifier(fulcioCert, ctfePublicKey))
+        .rekorClient(RekorClient.builder().build(), RekorVerifier.newRekorVerifier(rekorPublicKey))
+        .oidcClient(WebOidcClient.builder().build())
+        .signer(Signers.newEcdsaSigner());
+    return builder;
+  }
+
+  public static Builder builderForStaging()
```
patflynn, line 84: same naming concern here.

src/main/java/dev/sigstore/KeylessSigner.java
```diff
@@ -0,0 +1,185 @@
...
+    fulcioVerifier.verifyCertChain(signingCert);
+    // TODO: this signing workflow mandates SCTs, but fulcio itself doesn't, figure out a way to
+    // allow that to be known
+    fulcioVerifier.verifySct(signingCert);
+
+    // TODO: this reads the whole artifact which is probably expensive for large files, the signer
+    // interface should probably just sign non-digest artifacts instead of always applying sha256
+    // before signing. Then we can just apply the input stream to the sha256 calculator
+    // and send that digest to the signer
+    var artifactBytes = Files.readAllBytes(artifact);
```
patflynn, line 177: this should also be a stream shouldn't it?

src/main/java/dev/sigstore/KeylessSigner.java
patflynn: aha I've discovered your readme below. perhaps a pointer from here?

src/main/java/dev/sigstore/KeylessSigner.java
```diff
@@ -0,0 +1,185 @@
...
+      Preconditions.checkNotNull(rekorClient);
+      Preconditions.checkNotNull(rekorVerifier);
+      Preconditions.checkNotNull(oidcClient);
+      Preconditions.checkNotNull(signer);
+      return new KeylessSigner(
+          fulcioClient, fulcioVerifier, rekorClient, rekorVerifier, oidcClient, signer);
+    }
+  }
+
+  public void sign(Path artifact)
```
patflynn, line 156: should this return the rekor response?

src/main/java/dev/sigstore/KeylessSigner.java
loosebazooka: What do you mean? IF you mean inputstream then yes, I need to actually do the right thing here. That's the whole TODO.

src/main/java/dev/sigstore/KeylessSigner.java
loosebazooka: Should probably return a signature, a public key and a rekor response.

patflynn (approved):

PR #47, opened 2022-06-10
Use GSON as a singleton since config is shared across the SDK.

### LINKED ISSUE #42: Singleton the Gson object.
it's currently instantiated everywhere.
https://github.com/sigstore/sigstore-java/blob/7ad1722358b47651db516d39105672092c2286d0/src/main/java/dev/sigstore/json/GsonSupplier.java

patflynn:

Closes #42 

Supplier interface isn't really carrying any weight right now but left it since perhaps there's a world we might use DI in the SDK?

### DIFF
--- a/src/main/java/dev/sigstore/json/GsonSupplier.java
+++ b/src/main/java/dev/sigstore/json/GsonSupplier.java
@@ -27,14 +27,19 @@ import java.util.function.Supplier;
  * requests between sigstore and this client -- and should probably be used for any api call to
  * sigstore that expects JSON.
  */
-public class GsonSupplier implements Supplier<Gson> {
+public enum GsonSupplier implements Supplier<Gson> {
+  GSON;
+
+  private final Gson gson =
+      new GsonBuilder()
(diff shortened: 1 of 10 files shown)


src/main/java/dev/sigstore/json/GsonSupplier.java
```diff
@@ -27,14 +27,19 @@
  * requests between sigstore and this client -- and should probably be used for any api call to
  * sigstore that expects JSON.
  */
-public class GsonSupplier implements Supplier<Gson> {
+public enum GsonSupplier implements Supplier<Gson> {
```
loosebazooka, line 30: I don't know if https://immutables.github.io/immutable.html#singleton-instances matters, but they have singleton support.

loosebazooka (approved):

src/main/java/dev/sigstore/json/GsonSupplier.java
patflynn: GTK. I'm not sure about either pattern right now, so hopefully I'll have more of an opinion once I get back into the swing of things.

PR #50, opened 2022-06-10
add inclusion proof to Verification.

patflynn:

Adds inclusion proof data to our RekorEntry. Made it @nullable since it seems like inclusion proof is missing from the response to a post of a new entry to /entries.


src/main/java/dev/sigstore/rekor/client/RekorEntry.java
```diff
@@ -33,6 +35,36 @@ public interface RekorEntry {
   interface Verification {
     /** Return the signed entry timestamp. */
     String getSignedEntryTimestamp();
+
+    @Nullable
+    InclusionProof getInclusionProof();
```
loosebazooka, line 40: Could we just use optional instead of nullable though?

patflynn: doh. Still learning the habits. fixed.

loosebazooka (approved):

PR #52, opened 2022-06-15
Use common http parameters across http and grpc

loosebazooka:


patflynn:
Are you thinking of having http and grpc implementations over the long term?

patflynn:
oh I see, this is across the different clients.

patflynn (approved):

PR #64, opened 2022-07-15
Add verification for inclusion proof

loosebazooka:


src/main/java/dev/sigstore/rekor/client/RekorVerifier.java
```diff
@@ -90,6 +92,66 @@ public void verifyEntry(RekorEntry entry)
...
+                    new RekorVerificationException(
+                        "No inclusion proof was found in the rekor entry"));
+
+    var leafHash =
+        Hashing.sha256()
+            .hashBytes(combineBytes(new byte[] {0x00}, Base64.getDecoder().decode(entry.getBody())))
+            .asBytes();
+
+    // see: https://datatracker.ietf.org/doc/rfc9162/ section 2.1.3.2
+    var fn = inclusionProof.getLogIndex();
```
patflynn, line 111: I find the spec algorithm variable names pretty unintuitive
were you tempted to rename them for the java implementation to something more java like?
you could add a comment or something to match them to the spec names

src/main/java/dev/sigstore/rekor/client/RekorVerifier.java
```diff
@@ -90,6 +92,66 @@ public void verifyEntry(RekorEntry entry)
...
+    }
+
+    var calcuatedRootHash = Hex.toHexString(r);
+    if (!calcuatedRootHash.equals(inclusionProof.rootHash())) {
+      throw new RekorVerificationException(
+          "Calculated inclusion proof root hash does not match provided root hash\n"
+              + calcuatedRootHash
+              + "\n"
+              + inclusionProof.rootHash());
+    }
```
patflynn, line 141: should we be verifying the logid using the reko public key?

loosebazooka: yeah probs. I think I have a TODO somewhere else about that: https://github.com/sigstore/sigstore-java/blob/main/src/main/java/dev/sigstore/rekor/client/RekorVerifier.java#L69

src/main/java/dev/sigstore/rekor/client/RekorVerifier.java
loosebazooka: sgtm, I'll update

patflynn (approved):

Issue #34, opened 2022-06-01
Verify rekor logID against public key in verifier


loosebazooka:
I think we can check if the logId matches the calculated hash for the public key we are using. This means we can surface error messages likes "public key does not match entry logid" or something.

PR #71, opened 2022-08-08
Updates to rekor verification

loosebazooka:
- Verify logId matches supplied public key for verification
- Clean up some error messages
- Use json canonicalizer instead of custom logic

### DIFF
--- a/src/main/java/dev/sigstore/rekor/client/RekorVerifier.java
+++ b/src/main/java/dev/sigstore/rekor/client/RekorVerifier.java
@@ -15,23 +15,22 @@
  */
 package dev.sigstore.rekor.client;
 
-import static dev.sigstore.json.GsonSupplier.GSON;
-
 import com.google.common.hash.Hashing;
 import dev.sigstore.encryption.Keys;
 import dev.sigstore.encryption.signers.Verifiers;
 import java.io.IOException;
-import java.nio.charset.StandardCharsets;
 import java.security.*;
 import java.security.spec.InvalidKeySpecException;
 import java.util.Base64;
-import java.util.LinkedHashMap;
 import org.bouncycastle.util.encoders.Hex;
 
 /** Verifier for rekor entries. */
 public class RekorVerifier {
   private final PublicKey rekorPublicKey;
   private final String verifierAlgorithm;
+
+  // A calculated logId from the transparency log (rekor) public key
+  private final String calculatedLogId;
 
   public static RekorVerifier newRekorVerifier(byte[] rekorPublicKey)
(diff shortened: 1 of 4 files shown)


loosebazooka:
fixes #34

src/main/java/dev/sigstore/rekor/client/RekorVerifier.java
```diff
@@ -58,25 +55,16 @@ public void verifyEntry(RekorEntry entry)
...
-    signableContent.put("integratedTime", entry.getIntegratedTime());
-    signableContent.put("logID", entry.getLogID());
-    signableContent.put("logIndex", entry.getLogIndex());
-
-    // TODO: I think we can verify the logID (sha256 of log public key) here too
-    // to provide the user with some useful information
-    // (https://github.com/sigstore/sigstore-java/issues/34)
-
-    var signableJson = GSON.get().toJson(signableContent);
+    String calculatedLogId = Hashing.sha256().hashBytes(rekorPublicKey.getEncoded()).toString();
```
patflynn, line 58: this could be a field maybe?

patflynn:
I think this could become problematic once we have multiple tuf roots configurable, but we can tackle that once it's added. I'm still working out how it's supposed to work.

patflynn (approved):

loosebazooka:
> I think this could become problematic once we have multiple tuf roots configurable, but we can tackle that once it's added. I'm still working out how it's supposed to work.

Yeah I think it's a bit hard to visualize this now. The other thing we need to think about is signing multiple artifacts -- should help with the maven work.

src/main/java/dev/sigstore/rekor/client/RekorVerifier.java
```diff
@@ -44,6 +43,7 @@ public static RekorVerifier newRekorVerifier(byte[] rekorPublicKey)
   private RekorVerifier(PublicKey rekorPublicKey, String verifierAlgorithm) {
     this.rekorPublicKey = rekorPublicKey;
     this.verifierAlgorithm = verifierAlgorithm;
+    this.calculatedLogId = Hashing.sha256().hashBytes(rekorPublicKey.getEncoded()).toString();
```
patflynn, line 46: just FYI that Asra was telling me this probably won't work long term with changes coming down the pipe. I think it's fine for now.

patflynn (approved):

src/main/java/dev/sigstore/rekor/client/RekorVerifier.java
loosebazooka: makes sense. we can deal with that as it happens. It's not an essential check either.

PR #75, opened 2022-08-12
Allow signers and verifiers to work on digests

loosebazooka:
- Add concrete verifier types, package up verifier logic


loosebazooka:
just some stuff I ran into when writing the final keyless verification flows

patflynn:
could we just remove artifact bytes verification if digest based is the right way to go?

loosebazooka:
Signers are used in a few other places. Like oidc subject signing

src/main/java/dev/sigstore/encryption/signers/RsaVerifier.java
```diff
@@ -0,0 +1,52 @@
...
+    this.publicKey = publicKey;
+  }
+
+  @Override
+  public PublicKey getPublicKey() {
+    return publicKey;
+  }
+
+  @Override
+  public boolean verify(byte[] artifact, byte[] signature)
```
patflynn, line 35: these two methods look identical apart from the algo. maybe delegate to internal shared method?

loosebazooka: Man I had them all as parameterized versions of the same class. But then this weirdo rsa hash prefix got me good

patflynn (approved):

Issue #60, opened 2022-07-11
Pinned Fulcio and Rekor root certs should be updatable via TUF


patflynn:
Currently we have statically included the Rekor and Fulcio public keys into the library.  These keys should be updatable via TUF.

patflynn:
Here is a list of follow on items that I need to do:

- [ ] Handle upcoming addition of PEM headers to role public keys. (ex)[https://github.com/sigstore/sigstore/compare/main...asraa:sigstore:test-migrate-root?expand=1]
- [ ] Implement local store interface to replace direct filesystem read/write
- [ ] Implement snapshot update
- [ ] Implement timestamp update
- [ ] Implement target updates
- [ ] add support for creating/managing map.json. I suspect we want to isolate our map.json repos from other clients.
- [ ] API for fetching TUF targets (rekor and fulcio public keys) with meta support for service url -> root mapping.  See client design (here)[https://docs.google.com/document/d/1QWBvpwYxOy9njAmd8vpizNQpPti9rd5ugVhji0r3T4c/edit]

Obviously as we add more resource types we need to refactor.  The good news is most of the weirdness in the parsing and verification has been tackled so the rest of the resources should be pretty quick. (famous last words).

vlsi:
The question is how the default caching should work.
It would be weird if every signing would require TUF update.
On the other hand, if we claim `$HOME/.sigstore-java/caches/tuf`, then we would need to plan for concurrent access (e.g. multiple `sigstore-java` trying to update TUF concurrently).
On the other hand, it might be fun to have cross-ecosystem TUF cache, so the cache folder could be like `.sigstore/caches/tuf`

loosebazooka:
done!

PR #77, opened 2022-08-15
Add TUF model package

patflynn:
Uses GSON to import TUF metadata.

WIP towards #60.


src/main/java/dev/sigstore/json/GsonSupplier.java
```diff
@@ -30,13 +30,23 @@
...
-          .registerTypeAdapter(byte[].class, new GsonByteArrayAdapter())
-          .registerTypeAdapterFactory(new GsonAdaptersRekorEntry())
-          .registerTypeAdapterFactory(new GsonAdaptersRekorEntryBody())
-          .disableHtmlEscaping()
-          .create();
+  private final Gson gson;
+
+  GsonSupplier() {
+    GsonBuilder gsonBuilder = new GsonBuilder();
+    for (TypeAdapterFactory factory : ServiceLoader.load(TypeAdapterFactory.class)) {
```
loosebazooka, line 37: does this just autoload all the typeadapters in the classpath?

patflynn: yeah pretty much. Immutables registers them as META-INF/services/com.google.gson.TypeAdapterFactory, and then this code just loads them all up.

patflynn: https://immutables.github.io/json.html

patflynn: I've changed the code back to using explicit registration.

loosebazooka (approved):
I can't really tell if any types can be "empty" in which case I think they should be Optional<...>. But anyway, lgtm.

PR #78, opened 2022-08-15
Keyless Verifier implementation

loosebazooka:

~~The keyless verifier at the library level might need to pass on enough information for tooling(build, cli, etc) to provide good feedback to the user. So this is what I came up with. Not super sure about it though.~~

Switched it up, no collector anymore. Just a simple exception if things cannot be verified.

### DIFF
--- a/src/main/java/dev/sigstore/KeylessVerifier.java
+++ b/src/main/java/dev/sigstore/KeylessVerifier.java
@@ -1,0 +1,197 @@
+/*
+ * Copyright 2022 The Sigstore Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package dev.sigstore;
+
+import com.google.api.client.util.Preconditions;
+import com.google.common.io.Resources;
+import dev.sigstore.encryption.certificates.Certificates;
+import dev.sigstore.encryption.signers.Verifiers;
+import dev.sigstore.fulcio.client.FulcioVerificationException;
+import dev.sigstore.fulcio.client.FulcioVerifier;
+import dev.sigstore.fulcio.client.SigningCertificate;
+import dev.sigstore.rekor.client.*;
+import java.io.IOException;
+import java.net.URI;
+import java.security.*;
+import java.security.cert.CertificateException;
+import java.security.cert.CertificateExpiredException;
+import java.security.cert.CertificateNotYetValidException;
+import java.security.spec.InvalidKeySpecException;
+import java.sql.Date;
+import java.time.Instant;
+import java.util.Optional;
+
+/** Verify hashrekords from rekor signed using the keyless signing flow with fulcio certificates. */
+public class KeylessVerifier {
+  private final FulcioVerifier fulcioVerifier;
+  private final RekorVerifier rekorVerifier;
+  private final RekorClient rekorClient;
+
+  private KeylessVerifier(
+      FulcioVerifier fulcioVerifier, RekorClient rekorClient, RekorVerifier rekorVerifier) {
+    this.fulcioVerifier = fulcioVerifier;
+    this.rekorClient = rekorClient;
+    this.rekorVerifier = rekorVerifier;
+  }
+
+  public static KeylessVerifier.Builder builder() {
+    return new KeylessVerifier.Builder();
+  }
+
+  public static class Builder {
+    private FulcioVerifier fulcioVerifier;
+    private RekorClient rekorClient;
+    private RekorVerifier rekorVerifier;
+
+    public KeylessVerifier.Builder fulcioVerifier(FulcioVerifier fulcioVerifier) {
+      this.fulcioVerifier = fulcioVerifier;
+      return this;
+    }
+
+    public KeylessVerifier.Builder rekorClient(
+        RekorClient rekorClient, RekorVerifier rekorVerifier) {
+      this.rekorClient = rekorClient;
+      this.rekorVerifier = rekorVerifier;
+      return this;
+    }
+
+    public KeylessVerifier build() {
+      Preconditions.checkNotNull(fulcioVerifier);
+      Preconditions.checkNotNull(rekorVerifier);
+      Preconditions.checkNotNull(rekorClient);
+      return new KeylessVerifier(fulcioVerifier, rekorClient, rekorVerifier);
+    }
+
+    public Builder sigstorePublicDefaults()
+        throws IOException, InvalidAlgorithmParameterException, CertificateException,
+            InvalidKeySpecException, NoSuchAlgorithmException {
+      var fulcioCert =
+          Resources.toByteArray(
+              Resources.getResource("dev/sigstore/tuf/production/fulcio_v1.crt.pem"));
+      var ctfePublicKey =
+          Resources.toByteArray(Resources.getResource("dev/sigstore/tuf/production/ctfe.pub"));
+      var rekorPublicKey =
+          Resources.toByteArray(Resources.getResource("dev/sigstore/tuf/production/rekor.pub"));
+      fulcioVerifier(FulcioVerifier.newFulcioVerifier(fulcioCert, ctfePublicKey));
+      rekorClient(RekorClient.builder().build(), RekorVerifier.newRekorVerifier(rekorPublicKey));
+      return this;
+    }
+
+    public Builder sigstoreStagingDefaults()
+        throws IOException, InvalidAlgorithmParameterException, CertificateException,
+            InvalidKeySpecException, NoSuchAlgorithmException {
+      var fulcioCert =
+          Resources.toByteArray(Resources.getResource("dev/sigstore/tuf/staging/fulcio.crt.pem"));
+      var ctfePublicKey =
+          Resources.toByteArray(Resources.getResource("dev/sigstore/tuf/staging/ctfe.pub"));
+      var rekorPublicKey =
+          Resources.toByteArray(Resources.getResource("dev/sigstore/tuf/staging/rekor.pub"));
+      fulcioVerifier(FulcioVerifier.newFulcioVerifier(fulcioCert, ctfePublicKey));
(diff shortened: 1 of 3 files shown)


src/main/java/dev/sigstore/KeylessVerificationResult.java
```diff
@@ -0,0 +1,128 @@
...
+  }
+
+  public static class Builder {
+    private final List<Check> passingChecks = new ArrayList<>();
+
+    private Builder() {}
+
+    public Builder checkPassed(Check check) {
+      if (passingChecks.contains(check)) {
+        throw new IllegalStateException(
```
patflynn, line 72: If I'm understanding the API correctly, I think this is IllegalArgumentException, because the caller should not place the check into the same list twice.  If we used a Set here it would be a no-op.

src/main/java/dev/sigstore/KeylessVerificationResult.java
```diff
@@ -0,0 +1,128 @@
...
+  public static Builder builder() {
+    return new Builder();
+  }
+
+  public static class Builder {
+    private final List<Check> passingChecks = new ArrayList<>();
+
+    private Builder() {}
+
+    public Builder checkPassed(Check check) {
```
patflynn, line 70: maybe (set|add)Passed(Check check)?

src/main/java/dev/sigstore/KeylessVerificationResult.java
```diff
@@ -0,0 +1,128 @@
...
+
+  public Optional<String> getFailedReason() {
+    return Optional.ofNullable(failedReason);
+  }
+
+  public List<Check> getPassingChecks() {
+    return passingChecks;
+  }
+
+  public String generateJsonReport() {
```
patflynn, line 111: is this like the thing you see on the cosign CLI after you call verify successfully?

src/main/java/dev/sigstore/KeylessVerifier.java
```diff
@@ -0,0 +1,179 @@
...
+    this.fulcioVerifier = fulcioVerifier;
+    this.rekorClient = rekorClient;
+    this.rekorVerifier = rekorVerifier;
+  }
+
+  public static KeylessVerifier.Builder builder() {
+    return new KeylessVerifier.Builder();
+  }
+
+  public static class Builder {
```
patflynn, line 56: is the builder worth it here? Could a static factory method be simpler?

src/main/java/dev/sigstore/KeylessVerifier.java
```diff
@@ -0,0 +1,179 @@
...
+    var signingCert = SigningCertificate.from(Certificates.fromPemChain(certChain));
+    var leafCert = signingCert.getLeafCertificate();
+
+    var checks = KeylessVerificationResult.builder();
+
+    try {
+      fulcioVerifier.verifyCertChain(signingCert);
+      checks.checkPassed(Check.FULCIO_CERT);
+    } catch (FulcioVerificationException ex) {
+      return checks.checkFailed(Check.FULCIO_CERT, ex.getMessage());
```
patflynn, line 108: if we're not collecting errors I'm a little confused why we bother with this checks collection at all.

src/main/java/dev/sigstore/KeylessVerifier.java
```diff
@@ -0,0 +1,179 @@
...
+    }
+
+    var hashedRekordRequest =
+        HashedRekordRequest.newHashedRekordRequest(
+            artifactDigest, Certificates.toPemBytes(leafCert), signature);
+    Optional<RekorEntry> rekorEntry;
+    try {
+      rekorEntry = rekorClient.getEntry(hashedRekordRequest);
+    } catch (IOException ex) {
+      return checks.checkFailed(Check.REKOR_ENTRY_EXISTS, "retrieval failure: " + ex.getMessage());
```
patflynn, line 125: Technically this isn't about existence check but about a network failure or something.  Maybe just pass on to client?

src/main/java/dev/sigstore/KeylessVerifier.java
```diff
@@ -0,0 +1,179 @@
...
+    }
+    if (rekorEntry.isEmpty()) {
+      return checks.checkFailed(Check.REKOR_ENTRY_EXISTS, "entry not found");
+    }
+    checks.checkPassed(Check.REKOR_ENTRY_EXISTS);
+
+    if (!hashedRekordRequest
```
patflynn, line 132: this may be an overly defensive check :)

src/main/java/dev/sigstore/KeylessVerifier.java
```diff
@@ -0,0 +1,179 @@
...
+
+    try {
+      rekorVerifier.verifyEntry(rekorEntry.get());
+      checks.checkPassed(Check.REKOR_ENTRY_SIGNED);
+    } catch (RekorVerificationException ex) {
+      return checks.checkFailed(Check.REKOR_ENTRY_SIGNED, ex.getMessage());
+    }
+
+    try {
+      rekorVerifier.verifyInclusionProof(rekorEntry.get());
```
patflynn, line 152: Zach was mentioning that an inclusion proof is not something that clients should do by default, and that it's more for auditors of the log to make sure it's trustworthy.. Which reminds me we want a default verification method that doesn't hit Rekor.

patflynn (changes requested):

src/main/java/dev/sigstore/KeylessVerifier.java
```diff
@@ -0,0 +1,179 @@
...
+    public KeylessVerifier.Builder rekorClient(
+        RekorClient rekorClient, RekorVerifier rekorVerifier) {
+      this.rekorClient = rekorClient;
+      this.rekorVerifier = rekorVerifier;
+      return this;
+    }
+
+    public KeylessVerifier build() {
+      Preconditions.checkNotNull(fulcioVerifier);
+      Preconditions.checkNotNull(rekorVerifier);
```
loosebazooka, line 75: also need to check not null here on rekor client.

src/main/java/dev/sigstore/KeylessVerifier.java
loosebazooka: Yeah there would be a verifyOnline and verifyOffline (we probably need better naming too)

src/main/java/dev/sigstore/KeylessVerifier.java
loosebazooka: I mean this is what I was not feeling great about. Is there any point in collecting these things?

src/main/java/dev/sigstore/KeylessVerificationResult.java
loosebazooka: Yeah maybe the tooling that uses this library could generate a report for someone if something failed. I dunno

src/main/java/dev/sigstore/KeylessVerificationResult.java
```diff
@@ -0,0 +1,128 @@
...
+    this.failedCheck = failedCheck;
+    this.failedReason = failedReason;
+  }
+
+  public static Builder builder() {
+    return new Builder();
+  }
+
+  public static class Builder {
+    private final List<Check> passingChecks = new ArrayList<>();
```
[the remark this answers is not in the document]
loosebazooka: I thought order mattered initially, but I guess it doesn't.

src/main/java/dev/sigstore/KeylessVerificationResult.java
loosebazooka: I don't want it to be a no-op? Each check should only happen once. If there are multiple parts, it should be two checks.

src/main/java/dev/sigstore/KeylessVerificationResult.java
```diff
@@ -0,0 +1,128 @@
...
+        throw new IllegalStateException(
+            "Passing check: " + check + " is already in passing checks list");
+      }
+      passingChecks.add(check);
+      return this;
+    }
+
+    public KeylessVerificationResult checkFailed(Check check, String reason) {
+      if (passingChecks.contains(check)) {
+        throw new IllegalStateException(
```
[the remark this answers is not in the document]
loosebazooka: ya

src/main/java/dev/sigstore/KeylessVerifier.java
loosebazooka: image

but yeah, I'm taking it out

loosebazooka:
reworked it to just be a verification exception.

loosebazooka:
still gotta write tests though

src/main/java/dev/sigstore/KeylessVerifier.java
```diff
@@ -0,0 +1,197 @@
...
+   * Verify that the inputs can attest to the validity of a signature using sigstore's keyless
+   * infrastructure. If no exception is thrown, it should be assumed verification has passed.
+   *
+   * @param artifactDigest the sha256 digest of the artifact that was signed
+   * @param certChain the certificate chain obtained from a fulcio instance
+   * @param signature the signature on the artifact
+   * @throws KeylessVerificationException if the signing information could not be verified
+   * @throws IOException if an error occurred communicating with remote rekor
+   */
+  public void verifyOnline(byte[] artifactDigest, byte[] certChain, byte[] signature)
```
patflynn, line 120: Most of the formats I've seen so store cert and cert chain separately. https://docs.google.com/document/d/1gucjOA_bGyRjK6TeaOI-X5GIUv8WsPzeMDMkq25Kv4Y/edit is going to get implemented and also does so. Should we be accepting them as separate params?

src/main/java/dev/sigstore/KeylessVerifier.java
patflynn: I guess the question (probably a dumb one) I have is what is online verification for? What is the usecase?

src/main/java/dev/sigstore/KeylessVerifier.java
loosebazooka: hrmm... I dunno, we'd have to combine them together anyway during verification. I think we can accept them as separate params in the future as we work things out. But I think this is sufficient for now? I could be convinced otherwise.

src/main/java/dev/sigstore/KeylessVerifier.java
loosebazooka: I would assume a package repository would want to do it at upload time? If it's only useful during log verification and completely unnecessary for individual artifacts -- maybe there's no point.

Kinda waiting on the bundle offline spec to finalize before anything. I think doing this now doesn't preclude us from changing it later (we would be 0.1.0 for people to play with). We can save the harder decisions for when we start to get some feedback?

patflynn: SG

patflynn (approved):

PR #79, opened 2022-08-16
Use grpc client for fulcio

### LINKED ISSUE #65: Change Fulcio to call V2 API
**Description**

We'd like to deprecate the V1 API, so we should switch the call to `api/v1/signingCert` to `api/v2/signingCert`. You can optionally switch to calls over gRPC too, but if you want to not pull in gRPC, HTTP still works.

cc @loosebazooka

loosebazooka:
- FulcioGrpcClient -> FulcioClient
- Remove old FulcioClient
- Fix some test where I was checking nullable
  on an optional "oops face!"

fixes #65


src/main/java/dev/sigstore/fulcio/client/FulcioClient.java
```diff
@@ -86,43 +89,57 @@ public FulcioClient build() {
    *
    * @param cr certificate request parameters
    * @return a {@link SigningCertificate} from fulcio
-   * @throws IOException if the http request fials
```
patflynn, line 89: typo "fials"

src/main/java/dev/sigstore/fulcio/client/FulcioClient.java
```diff
@@ -86,43 +89,57 @@ public FulcioClient build() {
...
-    try (InputStream content = resp.getContent()) {
-      var signingCert =
-          SigningCertificate.newSigningCertificate(
-              CharStreams.toString(new InputStreamReader(content, resp.getContentCharset())),
-              sctHeader);
-      if (signingCert.getDetachedSct().isEmpty() && !signingCert.hasEmbeddedSct() && requireSct) {
-        throw new IOException(
-            "no signed certificate timestamps were found in response from Fulcio");
+      throws InterruptedException, CertificateException, IOException {
+    // TODO: we should reuse this message channel across various executions of this client and
```
patflynn, line 95: what's the impact of not doing this? should there be a bug?

src/main/java/dev/sigstore/fulcio/client/FulcioClient.java
```diff
@@ -86,43 +89,57 @@ public FulcioClient build() {
...
+              .build();
+
+      var certs = client.createSigningCertificate(req);
+
+      if (certs.getCertificateCase()
+          == dev.sigstore.fulcio.v2.SigningCertificate.CertificateCase
+              .SIGNED_CERTIFICATE_DETACHED_SCT) {
+        if (certs.getSignedCertificateDetachedSct().getSignedCertificateTimestamp().isEmpty()
+            && requireSct) {
+          throw new IOException(
```
patflynn, line 129: IOException is for unexpected network/file system errors isn't it?  This looks like serialization worked fine but for some reason the entry doesn't match expectations.  Perhaps FulcioVerificationException or just RuntimeException?

patflynn (approved):

src/main/java/dev/sigstore/fulcio/client/FulcioClient.java
loosebazooka: yeah I think we would need an autoclosable or something so users could do:
```
try (var connection = FulcioClient.newConnection()) {
  connection.signingCert()
  ...
  connection.differentSigningCert()
}
```
Right now it's a bit expensive, but I don't think dangerous.

loosebazooka: I don't actually know a reason why someone would want to keep the connection open for too long with fulcio. We probably only need one cert to sign a bunch of artifacts.

loosebazooka: updated the comment

src/main/java/dev/sigstore/fulcio/client/FulcioClient.java
loosebazooka: switched to CertificateException

patflynn (approved):

PR #88, opened 2022-08-29
Add release action

loosebazooka:
Manually triggered release action will tag and generate
java and slsa release artifacts.

Does not sign artifacts with sigstore/gpg yet.

### DIFF
--- a/.github/workflows/tag-and-build-release.yaml
+++ b/.github/workflows/tag-and-build-release.yaml
@@ -1,0 +1,108 @@
+name: Tag and Build Release
+on:
+  workflow_dispatch:
+    inputs:
+      release_version:
+        description: new release version
+        required: true
+        default: (for example, 0.1.0)
+
+jobs:
+  checks:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check inputs
+        run: |
+          if [[ ! "${{ github.event.inputs.release_version }}" =~ ^[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
+            echo 'version "${{ github.event.inputs.release_version }}" not in ###.###.### format'
+            exit 1
+          fi
+  ci:
+    uses: ./.github/workflows/ci.yaml
+
+  create-tag:
+    needs: [checks, ci]
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+    steps:
+      - name: tag
+        uses: actions/github-script@v5
+        with:
+          script: |
+            github.rest.git.createRef({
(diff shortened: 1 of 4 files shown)


.github/workflows/tag-and-build-release.yaml
```diff
@@ -0,0 +1,108 @@
...
+        required: true
+        default: (for example, 0.1.0)
+
+jobs:
+  checks:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check inputs
+        run: |
+          if [[ ! "${{ github.event.inputs.release_version }}" =~ ^[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
```
patflynn, line 16: you don't want to allow SNAPSHOT or ...-BETA, ..-RC1 type version releases?

loosebazooka:
oh goddamnit, looks like I'm spending the rest of the day fixing javadoc errors.

gradle.properties
```diff
@@ -1,2 +1,2 @@
 group=dev.sigstore
-version=0.1-SNAPSHOT
+version=dev-SNAPSHOT
```
patflynn, line 2: dev?

.github/workflows/tag-and-build-release.yaml
```diff
@@ -0,0 +1,108 @@
+name: Tag and Build Release
```
patflynn, line 1: I used to trigger releases by pushing tags, how does this one get triggered?

.github/workflows/tag-and-build-release.yaml
loosebazooka: we could, but I dunno, maybe it doesn't matter right now.

gradle.properties
loosebazooka: yeah I dunno, I was thinking we avoid the whole dance of constantly updating this version. We already have commits anyway, this number is much less relevant now.

.github/workflows/tag-and-build-release.yaml
loosebazooka: You just enter a version number like 0.0.0
This action tags the ref you're on (probably main), then builds it, and publishes the build artifacts to a new release.

It's one giant manually triggered pipeline.

loosebazooka: see: https://github.com/loosebazooka/slsa-java-test/actions/runs/2950354692

patflynn (dismissed):

patflynn (approved):

PR #92, opened 2022-08-31
Add local releaser script

loosebazooka:
- local releaser script queries for the latest release on github
- downloads all assets and signs with sigstore and gpg
- generates a maven central release bundle


gradle.properties
```diff
@@ -1,2 +1,2 @@
 group=dev.sigstore
-version=dev-SNAPSHOT
+version=SNAPSHOT
```
loosebazooka, line 2: this is intentional. dev was too weird as you pointed out.

scripts/sign_and_bundle_release.sh
```diff
@@ -0,0 +1,52 @@
...
+cd $RELEASE_DIR
+
+# cosign sign all the files
+echo "signing with cosign"
+for file in *; do
+  # skip intoto attestations, they are already signed
+  if [ $file == *.intoto.jsonl ] ; then
+    continue;
+  fi
+  COSIGN_EXPERIMENTAL=1 cosign sign-blob --yes $file --output-signature=$file.sig --output-certificate=$file.pem
```
patflynn, line 36: for now you should also include the bundle.

patflynn (changes requested):

scripts/sign_and_bundle_release.sh
loosebazooka: what bundle?

loosebazooka: oh the rekor bundle?

patflynn (approved):

PR #101, opened 2022-09-02
Add build-logic to make adding new modules easier

vlsi:
#### Summary

This moves sigstore-java sources to `sigstore-java` folder. The move itself is done in a separate commit to make review easier.

#### Release Note

* Added build-logic to make adding new modules easier

#### Documentation

WDYT?


vlsi:
Sorry for force-push storm. I just realized that I should move `/build.gradle.kts -> /sigstore-java/build.gradle.kts` before adding build logic.
Then Git realizes that `/build.gradle.kts` with added build logic is a brand new file, and it shows a better history for `/sigstore-java/build.gradle.kts`

build-logic/jvm/build.gradle.kts
```diff
@@ -0,0 +1,16 @@
...
+    id("build-logic.kotlin-dsl-gradle-plugin")
+}
+
+repositories {
+    gradlePluginPortal()
+}
+
+dependencies {
+    implementation(embeddedKotlinDsl())
+    implementation(plugin("com.github.vlsi.gradle-extensions", "1.82"))
```
vlsi, line 14: I guess this syntax would confuse dependabot/renovate, so I'll expand it

vlsi:
There's a recent idea to use `/gradle/plugins` folder instead of `/build-logic`: https://github.com/gradle/gradle/pull/21192

loosebazooka (commented):
I'm having a hard time following this, sorry to ask you to break this up again. I think what would make this easier for is to just create the new dir structure without build logic updates.
/sigstore-java/sigstore-java.

The build-logic seem to apply pretty strongly to a gradle plugin? But maybe I'm misunderstanding it.

We could do build logic updates in a new PR.

build-logic/publishing/src/main/kotlin/build-logic.publish-to-central.gradle.kts
```diff
@@ -0,0 +1,86 @@
...
+    val releaseDir = layout.buildDirectory.dir("release")
+    outputs.dir(releaseDir)
+
+    into(releaseDir)
+    rename("pom-default.xml", "${project.name}-${project.version}.pom")
+    rename("module.json", "${project.name}-${project.version}.module")
+}
+
+publishing {
+    publications.configureEach {
```
loosebazooka, line 76: I image this would have some unintended side effects if someone adds alternative publications (not that they would)?

build-logic/jvm/src/main/kotlin/build-logic.java.gradle.kts
```diff
@@ -0,0 +1,48 @@
...
+        from("$rootDir/LICENSE")
+        from("$rootDir/NOTICE")
+    }
+    manifest {
+        attributes["Bundle-License"] = "Apache-2.0"
+        attributes["Specification-Title"] = project.name + " " + project.description
+        attributes["Specification-Vendor"] = "dev.sigstore"
+        attributes["Implementation-Vendor"] = "dev.sigstore"
+        attributes["Implementation-Vendor-Id"] = "dev.sigstore"
+        // Implementation-Version is not here to make jar reproducible across versions
```
loosebazooka, line 46: does this really matter? Different versions of the jar don't necessarily need to be the same. Also I think we might need to include implementation-version to extract during runtime to set something like a user-agent string?

loosebazooka (approved):
lgtm, just a few comments.

build-logic/publishing/src/main/kotlin/build-logic.publish-to-central.gradle.kts
vlsi: I do not understand what you mean. Do you expect there will be a publication that is not a release artifact?

build-logic/jvm/src/main/kotlin/build-logic.java.gradle.kts
vlsi: It would be nice to have the same jar for both snapshot and non-snapshot build. For instance, I would like to have the same sha for 1.0.0-SNAPSHOT and 1.0.0 provided the code is the same. WDYT?

If you want to get the version in runtime, I incline to generating a simple class like SigstoreJavaVersion would be slightly better than storing version in the manifest.

I would expect that people would like to shade sigstore-java, and the manifest version is typically lost in that case, so manifest is not really reliable.

WDYT?

settings.gradle.kts
```diff
@@ -1 +1,6 @@
-rootProject.name = "sigstore-java"
+rootProject.name = "sigstore-java-root"
```
vlsi, line 1: I've no a good name for "the root project".

`sigstore-java` is taken by the java library itself, so it would probably be confusing to have `sigstore-java` under `sigstore-java` project. That is why I renamed to root to `sigstore-java-root`.

An alternative option would be just `sigstore`, however, it sounds too generic for me.

build-logic/jvm/src/main/kotlin/build-logic.java.gradle.kts
loosebazooka: Makes sense. Sgtm

settings.gradle.kts
loosebazooka: I don't care too much about this. It will never be exposed to users, only devs.

build-logic/publishing/src/main/kotlin/build-logic.publish-to-central.gradle.kts
loosebazooka: Yeah I was probably being overly concerned about something that isn't going to happen.

vlsi:
> I think what would make this easier for is to just create the new dir structure without build logic updates.
>The build-logic seem to apply pretty strongly to a gradle plugin? But maybe I'm misunderstanding it.

I must have missed this.

I believe this PR adds only the bits that are required to build `sigstore-java`, and the build logic updates for building Gradle plugin are not here (they go in the last commit of https://github.com/sigstore/sigstore-java/pull/98)

loosebazooka (approved):

loosebazooka:
If this is ready to go, I'll merge it?

vlsi:
Go ahead

PR #114, opened 2022-09-03
Move gha-oidc-tests to a separate workflow that is run repo branches only

vlsi:
#### Summary

Move gha-oidc-tests to a separate workflow.

#### Release Note

NONE

#### Documentation

NONE


loosebazooka (approved):

loosebazooka:
I think what you proposed earlier makes the "status" check easier to configure: #113  ? wdyt?

vlsi:
I think having it in a separate file makes it easier to debug and configure.
I spent quite some time fighting with the expression in #113

PR #136, opened 2022-09-11
Update renovate.json

loosebazooka:
Add some groupings to renovate bot, so things update together.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -3,5 +3,15 @@
   "extends": [
     "config:base"
   ],
-  "schedule": ["every weekend"]
+  "schedule": ["every weekend"],
+  "packageRules": [
+    {
+      "matchPackagePrefixes": ["org.immutables"],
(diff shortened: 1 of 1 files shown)


renovate.json
```diff
@@ -3,5 +3,15 @@
   "extends": [
     "config:base"
   ],
-  "schedule": ["every weekend"]
+  "schedule": ["every weekend"],
+  "packageRules": [
+    {
+      "matchPackagePrefixes": ["org.immutables"],
```
vlsi, line 9: 2.9.2 has bom, so we can use it instead.
However, sigstore-java is not compatible with 2.9.2

vlsi (approved):
Makes sense, however, I would go for bom for immutables

loosebazooka:
Oh interesting, I'll switch to bom in a follow-up

PR #144, opened 2022-09-12
Remove conditional use of bouncy castle 

patflynn:
TUF code requires Bouncy Castle in all jdk versions

### DIFF
--- a/sigstore-java/src/main/java/dev/sigstore/encryption/Keys.java
+++ b/sigstore-java/src/main/java/dev/sigstore/encryption/Keys.java
@@ -46,23 +46,7 @@ public class Keys {
   private static final Logger log = Logger.getLogger(Keys.class.getName());
 
   static {
-    // Added for EdDSA support for Java <15
-
-    // This should work as JDK version strings are of the form '1.x.x' up to Java 8, and '9.x..'
-    // afterwards.
-    if (getJavaVersion() < 15) {
-      try {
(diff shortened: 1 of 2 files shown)


vlsi:
Just wondering: can tuf be located in its own module?

E.g.:
* tuf
* fulcio-client
* rekor-client
* verifier
* ...

patflynn:
I think it could be eventually. Currently it's very much a sigstore specific implementation.

loosebazooka (approved):

vlsi:
>I have no objection to breaking them out as modules. I'm not in any personal rush to do it. :)

Well, I see what you mean.

On the other hand, it might be useful to have "offline-no-dependency-verifier" module :)

patflynn:
Oh do you mean an offline signature verification mode?

vlsi:
>Oh do you mean an offline signature verification mode?

Exactly.
If I understand it right, offline signature verification can do without many third-party dependencies, and a "small" Java library might be helpful for the users to perform the verification.

patflynn:
I guess it depends on the use-case and just how off-line we want to be.  I think we want verification to not generate any rekor traffic by default, but that still requires parsing and verifying the signing bundle and tuf metadata.  It'd be a really good idea to specify an offline flow but that will be a lot easier to do after the new bundle spec lands and after we've added the TUF root data to the rekor entry.

PR #146, opened 2022-09-13
add unit testing for tuf key verification.

patflynn:
Required refactoring static verifiers factory method.  I normally have a good idea how to handle DI on the server-side, but I'm just making it up as I go here. I'm very open to changes. Eventually it might make sense to collapse the two levels of indirection between verification supplier and verification. 

Work towards #60 


loosebazooka (commented):
Just to start a discussion. I think I would just add the functional interface to Verifiers

```java
  @FunctionalInterface
  public interface Supplier {
      Verifier newVerifier(PublicKey publicKey) throws NoSuchAlgorithmException;
  }
```

on a method signature like
```java
/* package private */ TufClient(Verifiers.Supplier supplier)
```
you can inject with the method reference
```java
new TufClient(Verifiers::newVerifier)
```

If we don't need to expose Verifiers.Supplier we can define that functional interface in the TUF client itself.

I might also make the TUF client have one public static factory method `newTufClient` and have the various constructors you need for testing be package private?

patflynn:
Massive improvement! Now we're cooking with gas!

sigstore-java/src/main/java/dev/sigstore/tuf/TufClient.java
```diff
@@ -55,7 +55,21 @@ public class TufClient {
...
+  private Verifiers.Supplier verifiers;
+
+  public TufClient(Clock clock, Verifiers.Supplier verifiers) {
+    this.clock = clock;
+    this.verifiers = verifiers;
+  }
+
+  public TufClient(Verifiers.Supplier verifiers) {
+    this(Clock.systemUTC(), verifiers);
+  }
```
loosebazooka, line 68: should these be public?

loosebazooka: They can be package private and still be available for tests?

loosebazooka (approved):
Minor stuff but LGTM.

sigstore-java/src/main/java/dev/sigstore/tuf/TufClient.java
patflynn: I'll remove public. FYI the TufClient class at the moment is really just a giant place holder for the update logic. It's going to be refactored, and the current API is not reflective in any way of what the client API will be.

loosebazooka (approved):

PR #147, opened 2022-09-14
Use @EnabledIfOidcExists instead of several Gradle test tasks

vlsi:
#### Summary

The idea is to keep a single `test` task only, which would make IDE UI easier: it won't ask "which of test, testGithubOidc, testManual tasks would you like to use to execute the test".

#### Release Note

NONE

#### Documentation

NONE


vlsi:
A side-effect is that `./gradlew build` would attempt executing test with `oidcprovider in (ANY, MANUAL)`.
I'm inclined to add a property like `skipOidc` to allow skipping OIDC flows when testing on dev machine.

loosebazooka:
LGTM, but I would like a way to disable manual testing locally (calling to sign in)

vlsi:
>  disable manual testing locally

Run with `-PskipOidc=true` or add `skipOidc=true` to your `$HOME/.gradle/gradle.properties`.
Is that good enough?

PS https://docs.gradle.org/current/userguide/build_environment.html#sec:gradle_configuration_properties
PPS https://docs.gradle.org/current/userguide/build_environment.html#sec:project_properties

loosebazooka (approved):
Do our actions need an update?

loosebazooka:
So this will work because: `id-token:write` will never be accessible to an external PR? But those test just wont run anyway because of these checks?

.github/workflows/ci.yaml
```diff
@@ -31,6 +31,8 @@ jobs:
       fail-fast: false
 
     runs-on: ubuntu-latest
+    permissions:
+      contents: write
```
loosebazooka, line 35: should this be id-token:write?

vlsi: indeed

loosebazooka (approved):

loosebazooka:
> I'm inclined to add id-token: write to the regular ci.yaml

Id tokens are scoped to workflow, so there appears to be reduced risk is leaked.

PR #148, opened 2022-09-14
Add sandbox for Gradle samples

vlsi:
#### Summary

The sandbox provides "better" end-to-end tests as they use less "test infra" than tests we have in JUnit.
At the same time, it would help users to see how things could be used and wired.

#### Release Note

NONE

#### Documentation

NONE


loosebazooka:
I don't know that I fully understand what's going on here:
1. We are testing against release versions?
2. Is this more a samples directory than a sandbox?
3. Should this only execute on changes to sandbox dir?

vlsi:
>We are testing against release versions?

Not yet, however, it could be done with a very small change.

>Is this more a samples directory than a sandbox?

Quite right.

>Should this only execute on changes to sandbox dir?

Not really.

For instance, I recently learned that the dependencies between "plugin" modules should be `api`, not `implementation (see https://github.com/allure-framework/allure-gradle/pull/97 )
The issue does not surface when testing the plugin via TestKit (probably because it adds just "classes" to the classpath rather than testing against the build jar files).

loosebazooka (approved):

PR #149, opened 2022-09-14
Adds a Tuf local store abstraction and a file based implementation

patflynn:
TUF local repo store sets us up to be able to use an in-memory implementation for read-only file system scenarios.

Progress towards #60 

### DIFF
--- a/sigstore-testkit/src/main/java/dev/sigstore/testkit/tuf/TestResources.java
+++ b/sigstore-testkit/src/main/java/dev/sigstore/testkit/tuf/TestResources.java
@@ -1,0 +1,43 @@
+/*
+ * Copyright 2022 The Sigstore Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package dev.sigstore.testkit.tuf;
+
+import com.google.common.io.Resources;
+import dev.sigstore.json.GsonSupplier;
+import dev.sigstore.tuf.model.Root;
+
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+
+public class TestResources {
+
+  public static final Path CLIENT_TRUSTED_ROOT = Path.of(Resources.getResource("dev/sigstore/tuf/trusted-root.json").getPath());
+  public static final Path TUF_TEST_DATA_DIRECTORY = CLIENT_TRUSTED_ROOT.getParent();
+
+  public static void setupRepoFiles(String repoName, Path destinationDir, String... files)
+      throws IOException {
+    for (String file : files) {
+      Files.copy(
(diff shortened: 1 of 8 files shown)


vlsi:
@patflynn , what do you think of putting "abstractions" into `sigstore-testkit` folder?

They might be useful across more than a single module, and I'm inlined it might be slightly better if `test/src` includes code with `@Test` methods "only".

patflynn:
@vlsi You're suggesting putting the TestResources into testkit? I like the idea. This was the rule at Google too.  It's  a bit tricky in this case because I would have to move the associated resource data, which I like a little less from a navigatability standpoint for test maintainers.

vlsi:
> It's a bit tricky in this case because I would have to move the associated resource data

That is right, however, reading files from `src/...` might be not the best idea.

WDYT of using `Class.getResource(...)`?
Then `TestResources` could take `Class<>` argument, and it could use that to lookup the resources.
An alternative option is to use `com.google.common.io.Resources#getResource(java.lang.Class<?>, java.lang.String)`

patflynn:
Yeah that makes sense. I was avoiding it initially because the API itself doesn't read resources from the class path as the local repo and trusted root can be anywhere on the filesystem, but I should be able to bridge that.

patflynn:
This does get a little nasty though as circular deps are going to rare their ugly heads :).  These test support classes will be quite limited since they won't be able to use any sigstore-java types unless we break the tests out into a separate module..

vlsi:
> These test support classes will be quite limited since they won't be able to use any sigstore-java types

I do not follow you here. sigstore-testkit dependency on sigstore-java is just fine.

patflynn:
Aha! You're right. That's great! I guess Gradle doesn't include the test deps when testing module cycles. GTK.

vlsi:
>I guess Gradle doesn't include the test deps when testing module cycles. GTK.

Depending on "test" artifacts is weird, and Gradle does not provide out-of-the-box to "depend on files from src/test/..."

Here's a sample where `calcite-testkit` depends on `calcite-core`: https://github.com/apache/calcite/blob/e41555f568510df04d3883516b2195393a93426d/testkit/build.gradle.kts#L22, and it works great.

sigstore-testkit/src/main/java/dev/sigstore/testkit/tuf/TestResources.java
```diff
@@ -0,0 +1,43 @@
...
+import dev.sigstore.json.GsonSupplier;
+import dev.sigstore.tuf.model.Root;
+
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+
+public class TestResources {
+
+  public static final Path CLIENT_TRUSTED_ROOT = Path.of(Resources.getResource("dev/sigstore/tuf/trusted-root.json").getPath());
```
vlsi, line 28: I wonder if that works in case the resource comes from a jar

patflynn: :shrug: Should we cross that bridge when/if we get to it?

vlsi: I'm fine if it passes tests. I just got bitten multiple times by the fact that "resource loaded from jar does not have proper file path"

patflynn: I hear you. In this case it's just test code I think we're good. (famous last words) 

I think you need to click the approve button.

vlsi (approved):

patflynn:
Do you mean move the TestResources class to testkit as TufTestResources or
something like that?

PR #158, opened 2022-09-17
Add Gradle precompiled script plugin sample

vlsi:
#### Summary

The sample checks if `dev.sigstore.sign` works in a precompiled script plugin.
See https://docs.gradle.org/current/userguide/custom_plugins.html#sec:precompiled_plugins

#### Release Note

NONE

#### Documentation

NONE


loosebazooka (approved):
I'm okay with this, but I'm not convinced it's strictly necessary to document this. This just seems like how to do script plugins in general? But I guess it adds more to the sandbox test cases?

vlsi:
>This just seems like how to do script plugins in general?

exactly.

Then it is the way to test and ensure that dependency on `project(":sigstore-gradle:sigstore-gradle-sign-base-plugin")` is `api`, not `implementation`.

PR #160, opened 2022-09-21
refactor tuf file system store

patflynn:
My initial implementation had too much going on in the static factory, and creates uncertainty by having a constructor with a store path AND a trusted root.  I think it's better to initialize the store with the directory, and set the root afterwards. It also makes more sense to load these resources on demand as state can't get wonky and we'll pick up changes from other clients. 

Progress towards #60 


vlsi (approved):

patflynn:
Thanks! @vlsi

PR #162, opened 2022-09-22
Adds a convenience method for verifying delegated roles

patflynn:
The lower level method needs to stay for both testing and verifying sub-delegates that can occur in targets.json.


patflynn:
woops missed some changes. closing and will  retry.

PR #172, opened 2022-09-26
Multi artifact keyless signing of digests

loosebazooka:

There are some minor kotlin changes in here to reflect the new api.

### DIFF
--- a/sigstore-gradle/sigstore-gradle-sign-base-plugin/src/main/kotlin/dev/sigstore/sign/work/SignWorkAction.kt
+++ b/sigstore-gradle/sigstore-gradle-sign-base-plugin/src/main/kotlin/dev/sigstore/sign/work/SignWorkAction.kt
@@ -52,7 +52,7 @@ abstract class SignWorkAction : WorkAction<SignWorkParameters> {
             oidcClient(parameters.oidcClient.get().build() as OidcClient)
         }.build()
 
-        val result = signer.sign(inputFile.toPath())
+        val result = signer.signFile(inputFile.toPath())
         val signature = SigstoreBundle(
             mediaType = BundleMediaTypes.V1_JSON.value,
             timestampProof = RekorEntry(
--- a/sigstore-java/src/main/java/dev/sigstore/KeylessSigner.java
+++ b/sigstore-java/src/main/java/dev/sigstore/KeylessSigner.java
@@ -16,6 +16,8 @@
 package dev.sigstore;
 
 import com.google.api.client.util.Preconditions;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
 import com.google.common.hash.Hashing;
 import dev.sigstore.encryption.certificates.Certificates;
 import dev.sigstore.encryption.signers.Signer;
@@ -24,10 +26,7 @@ import dev.sigstore.fulcio.client.*;
 import dev.sigstore.oidc.client.OidcClient;
 import dev.sigstore.oidc.client.OidcException;
 import dev.sigstore.oidc.client.WebOidcClient;
-import dev.sigstore.rekor.client.HashedRekordRequest;
-import dev.sigstore.rekor.client.RekorClient;
-import dev.sigstore.rekor.client.RekorVerificationException;
-import dev.sigstore.rekor.client.RekorVerifier;
+import dev.sigstore.rekor.client.*;
 import java.io.IOException;
 import java.net.URI;
 import java.nio.charset.StandardCharsets;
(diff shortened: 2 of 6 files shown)


sigstore-gradle/sigstore-gradle-sign-base-plugin/src/main/kotlin/dev/sigstore/sign/work/SignWorkAction.kt
```diff
@@ -52,7 +52,8 @@ abstract class SignWorkAction : WorkAction<SignWorkParameters> {
             oidcClient(parameters.oidcClient.get().build() as OidcClient)
         }.build()
 
-        val result = signer.sign(inputFile.toPath())
+        val results = signer.sign(*KeylessSigner.asDigests(inputFile.toPath()))
+        val result = results.get(0);
```
vlsi, line 56:
```suggestion
        val result = results.first();
```

sigstore-java/src/main/java/dev/sigstore/KeylessSigner.java
```diff
@@ -152,10 +151,49 @@ public Builder sigstoreStagingDefaults()
...
 
-  public KeylessSigningResult sign(Path artifact)
+  /**
+   * Convenience method to convert files to their digests
+   *
+   * @param artifacts paths to files to convert
+   * @return an array of digests in the same order as the entered artifacts
+   * @throws IOException if issues occur processing the file contents
+   */
+  public static byte[][] asDigests(Path... artifacts) throws IOException {
```
vlsi, line 161: Do we have a class for `Hash`? (e.g. algorithm + bytes)
Could this be `List<Hash>`?

sigstore-gradle/sigstore-gradle-sign-base-plugin/src/main/kotlin/dev/sigstore/sign/work/SignWorkAction.kt
vlsi, line 56:
```suggestion
        val result = signer.sign(*KeylessSigner.asDigests(inputFile.toPath())).first()
```

vlsi: However, currently we still create keylesssigner request per file, so it would be nice to batch signing requests automatically.

In other words, add stateful API like in
* `create signing session`: remember URLs, and other settings.
* `sign artifact`:  It should perform OIDC auth if missing, or it could just reuse already existing keys if they are still valid.
* `close signing session`: throw session keys away

sigstore-java/src/main/java/dev/sigstore/KeylessSigner.java
loosebazooka: I had the varargs earlier because I was using multiple methods with similar signatures -- where lists wouldn't work. I think this can all be lists though. Not quite sure about `Hash` yet, lemme see.

vlsi: I see why varargs might be helpful, so we might consider both `asDigests(Path... artifacts)` and `asDigests(Iterable<? extends Path> artifacts)`. However, my question here was regarding the method result type: use `List` instead of array.

vlsi:
CI failure is like

```
  Execution failed for task ':signFile'.
  > There was a failure while executing work items
     > A failure occurred while executing dev.sigstore.sign.work.SignWorkAction
        > 'byte[][] dev.sigstore.KeylessSigner.asDigests(java.nio.file.Path[])'
```

It might be that `SignWorkAction` uses `sigstore-java` from Central, so we might need to adjust it so it picks the currently built jar.

sigstore-java/src/main/java/dev/sigstore/KeylessSigner.java
loosebazooka: yeah it made it easier to just drop it into the varargs sign...

loosebazooka:
This is failing on #188 for now, but ready for review.

sigstore-java/src/main/java/dev/sigstore/KeylessSigner.java
```diff
@@ -169,21 +186,77 @@ public KeylessSigningResult sign(Path artifact)
...
+    return result.build();
+  }
+
+  /**
+   * Convenience wrapper around {@link #sign(List)} to sign a single digest
+   *
+   * @param artifactDigest sha256 digest of the artifact to sign.
+   * @return a keyless singing results.
+   */
+  public KeylessSigningResult sign(byte[] artifactDigest)
```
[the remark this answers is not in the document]
loosebazooka: It does feel weird, but if the batch calls it in a loop, then it will have to call out to fulcio/dex multiple times too.

patflynn (dismissed):

loosebazooka:
@vlsi I'm a bit confused about how the gradle plugin is pulling in the sigstore-java dep. It appears to not be able to handle breaking changes here?

patflynn (approved):

PR #185, opened 2022-10-17
Update protobuf plugin

loosebazooka:
Looks like the new protobuf plugin handles a few more things that the old one. Updates to remove unnecessary config that was breaking things.

### DIFF
--- a/sigstore-java/build.gradle.kts
+++ b/sigstore-java/build.gradle.kts
@@ -1,15 +1,10 @@
-import com.google.protobuf.gradle.generateProtoTasks
 import com.google.protobuf.gradle.id
-import com.google.protobuf.gradle.ofSourceSet
-import com.google.protobuf.gradle.plugins
-import com.google.protobuf.gradle.protobuf
-import com.google.protobuf.gradle.protoc
 
 plugins {
     id("build-logic.java-published-library")
     id("com.diffplug.spotless") version "6.11.0"
     id("org.jsonschema2dataclass") version "4.4.0"
-    id("com.google.protobuf") version "0.8.19"
+    id("com.google.protobuf") version "0.9.1"
 }
 
 description = "A Java client for signing and verifying using Sigstore"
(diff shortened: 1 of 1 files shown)


patflynn (approved):

loosebazooka:
I think staging is going through something, I'll wait for it come back and run these tests again.

sigstore-java/build.gradle.kts
```diff
@@ -67,7 +63,7 @@ protobuf {
     generateProtoTasks {
         ofSourceSet("main").forEach() {
             it.plugins {
-                id("grpc")
+                id("grpc") {}
             }
         }
```
vlsi, line 68: Please use the following here:

```kotlin
    generateProtoTasks {
        ofSourceSet("main").configureEach {
            plugins {
                id("grpc")
            }
        }
    }
```

vlsi:
@loosebazooka , thanks.
Please replace `dependsOn("generateProto", "generateJsonSchema2DataClass0")` with `dependsOn("generateJsonSchema2DataClass0")` at the end of `/sigstore-java/build.gradle.kts`.
It looks like `dependsOn("generateProto")` is no longer needed.

sigstore-java/build.gradle.kts
loosebazooka: Oops must've left that in there while experimenting

vlsi: I would suggest replacing `forEach()` with `configureEach`, so `it.` is not needed before `plugins`

loosebazooka: Damn, I missed that whole suggestion

vlsi: I logged an issue for streamlining protobuf-gradle-plugin configuration: https://github.com/google/protobuf-gradle-plugin/issues/628

vlsi (approved):

loosebazooka:
@vlsi thanks for filing those protobuf gradle plugin bugs, I was also gonna see if they would update their kotlin-dsl example.

PR #186, opened 2022-10-19
Fix unused and redundant deps

loosebazooka:


vlsi (approved):

vlsi:
By the way, how does that even compile?

I think dependency on `sigstore-teskit` was needed for `@EnabledIfOidcExists` annotation which is declared in `sigstore-testkit`:
https://github.com/sigstore/sigstore-java/blob/80c21e50be4a73e2771503562a89c69c7129c8ae/sigstore-java/src/test/java/dev/sigstore/KeylessTest.java#L55

PR #190, opened 2022-10-21
Support multiple ctfe keys

### LINKED ISSUE #188: ctfe key rotated, we need to handle this
sct's can be signed by multiple instances of the ctfe log. Allow the verifier to use multiple keys.

loosebazooka:

This is kind of a temporary solution till we get tuf ready. I still think we should make some sort of "verification material provider" interface that our TUF client implements and maybe there are other implementations?. But... this fixes #188 for now.


patflynn:
my plan was to expose a simple lookup method per client that returns the public key for a specific URL. e.g. getPublicKey(URL).

loosebazooka:
yeah sgtm, we can refactor for that when ready.

sigstore-java/src/main/java/dev/sigstore/VerificationMaterial.java
```diff
@@ -0,0 +1,66 @@
...
+  }
+
+  /** Verification material for *.sigstage.dev */
+  public static class Staging {
+    public static byte[] fulioCert() throws IOException {
+      return Resources.toByteArray(
+          Resources.getResource("dev/sigstore/tuf/staging/fulcio.crt.pem"));
+    }
+
+    public static List<byte[]> ctfePublicKeys() throws IOException {
```
patflynn, line 52: I know this is temporary code but why not just parse the pub keys right here?

sigstore-java/src/main/java/dev/sigstore/fulcio/client/FulcioVerifier.java
```diff
@@ -59,15 +60,26 @@ public static FulcioVerifier newFulcioVerifier(byte[] fulcioRoot, byte @Nullable
...
+  private FulcioVerifier(TrustAnchor fulcioRoot, @Nullable List<PublicKey> ctfePublicKeys) {
     this.fulcioRoot = fulcioRoot;
-    if (ctfePublicKey != null) {
-      var ctLogInfo = new CTLogInfo(ctfePublicKey, "fulcio ct log", "unused-url");
+    if (ctfePublicKeys != null) {
+      var logInfos = new ArrayList<CTLogInfo>();
+      for (var pk : ctfePublicKeys) {
+        var ctLogInfo = new CTLogInfo(pk, "fulcio ct log", "unused-url");
+        logInfos.add(ctLogInfo);
+        System.out.println(Base64.getEncoder().encodeToString(ctLogInfo.getID()));
```
patflynn, line 73: do you mean to include this system out call?

sigstore-java/src/main/resources/dev/sigstore/tuf/README.md
```diff
@@ -13,9 +13,9 @@ We keep copies of the remote tuf repositories locally in
    for interfacing with *.sigstage.dev
 
 For this client to function we need the following keys
-1. CTFE public key (`ctfe.pub` \
+1. CTFE public keys (`ctfe.pub`, `ctfe_2022.pub`) \
```
patflynn, line 16: why do we need the old key?

loosebazooka: you need the old key to verify that older SCTs embedded into older certs can still be verified (since they are not re-issued).

sigstore-java/src/main/java/dev/sigstore/fulcio/client/FulcioVerifier.java
loosebazooka: yup... nope

sigstore-java/src/main/resources/dev/sigstore/tuf/README.md
loosebazooka: also I need to update htis readme

patflynn: hmm, I guess we need the full history to verify old rekor entries with SCTs that were created with older keys.. I need to talk to Asra I think because I thought she said it was fine to just bundle the latest in the client.

loosebazooka: we need a full key history for all of sigstore (unless there is an expiry date for software?)

loosebazooka: Yeaah there's still a bit more to do here. We need to allow the client to accept multiple fulcioRoots and rekorPubKeys.

sigstore-java/src/main/java/dev/sigstore/VerificationMaterial.java
loosebazooka: oh parse them into keys? I dunno, I guess we could stop sending `byte[]`s around.

loosebazooka: it might be weird for each implementation to have to parse it?

loosebazooka: lemme handle this is a followup. We'll return certs, and keys instead of just bytes everywhere.

loosebazooka: #191

loosebazooka:
@patflynn can we get this in and then followup, it'll fix the broken build.

patflynn (approved):

haydentherapper:
@loosebazooka We also need to add the new CTFE [prod key](https://github.com/sigstore/root-signing/blob/main/targets/ctfe_2022.pub), correct?

loosebazooka:
Correct. Will add in soon.

PR #200, opened 2022-10-23
implement TUF timestamp resource update

patflynn:
This PR adds a timestamp.json update flow. 

WIP towards #60 

FYI Once snapshot and target resource updates are implemented I intend to rename the TufClient class to something like 'Updater' which will primarily be an internal API.  The public API for this client is not reflected in the current implementation.


sigstore-java/src/main/java/dev/sigstore/json/GsonSupplier.java
```diff
@@ -43,18 +43,20 @@ public enum GsonSupplier implements Supplier<Gson> {
...
+          .registerTypeAdapterFactory(new GsonAdaptersTimestamp())
+          .registerTypeAdapterFactory(new GsonAdaptersTimestampMeta())
           .registerTypeAdapterFactory(new GsonAdaptersDelegationRole())
           .registerTypeAdapterFactory(new GsonAdaptersDelegations())
-          .registerTypeAdapterFactory(new GsonAdaptersRootRole())
           .registerTypeAdapterFactory(new GsonAdaptersSignature())
           .registerTypeAdapterFactory(new GsonAdaptersKey())
-          .registerTypeAdapterFactory(new GsonAdaptersRoot())
-          .registerTypeAdapterFactory(new GsonAdaptersTargets())
           .disableHtmlEscaping()
```
vlsi, line 60: Can the section be sorted alphabetically?

patflynn:
sure. done.

sigstore-java/src/test/java/dev/sigstore/tuf/TufClientTest.java
```diff
@@ -146,6 +148,97 @@ public void testRootUpdate_metaFileTooBig()
...
+
+  @Test
+  public void testTimestampUpdate_throwsSignatureVerificationException()
+      throws IOException, NoSuchAlgorithmException, InvalidKeySpecException, InvalidKeyException {
+    setupMirror("remote-timestamp-unsigned", "2.root.json", "3.root.json", "timestamp.json");
+    var client = createTimeStaticTufClient(localStore);
+    try {
+      client.updateRoot();
+      client.updateTimestamp();
+      fail();
```
vlsi, line 170: Frankly speaking, it would be nice to have `assertThrows` in such cases.
The bad thing with standalone `fail()` is that the test failure would be extremely obscure to a new developer. They would have to reverse-engineer the reasoning behind the test code.

If you add a relevant message, then the test failure would look like a decent bug report:

```java
assertThrows(
    SignatureVerificationException.class,
    () -> { test code },
    () -> "root signature is missing in ..., so updateRoot should fail"
);
```

The same goes for `assertEquals(0, e.getVerifiedSignatures());`
It produces unclear error messages.

patflynn:
looks like something about the oidc test is unhappy. seems unrelated to the PR. @loosebazooka FYI

sigstore-java/src/test/java/dev/sigstore/tuf/TufClientTest.java
```diff
@@ -146,6 +148,97 @@ public void testRootUpdate_metaFileTooBig()
...
+    } catch (SignatureVerificationException e) {
+      assertEquals(0, e.getVerifiedSignatures());
+      assertEquals(1, e.getRequiredSignatures());
```
vlsi, line 173: You might want to consider https://approvaltests.com/ for verifications like that. I do not insist on approvaltests, however, it might help us to mass-update the expected outcomes should the output format change.

sigstore-java/src/test/java/dev/sigstore/tuf/TufClientTest.java
patflynn: Hey Vladimir,

I agree with you there's room for improvement here. I forgot to add error messages in these cases so I've gone and added them. I can't say I'm a fan of the assertThrows approach for cases where I want to do extra verification of the exception contents. It just seems to make the code more deeply nested and harder to read.  I hope the messages I've included improve things somewhat.

sigstore-java/src/test/java/dev/sigstore/tuf/TufClientTest.java
patflynn: Wow that looks super interesting.  I don't want to learn and pull in a new testing framework for this case, but I'm very much open to taking a look at this and how it can improve these tests in the future.  If you'd like to send a demo PR so that we can see how it works I'd be 100% for it.

sigstore-java/src/test/java/dev/sigstore/tuf/TufClientTest.java
vlsi: > It just seems to make the code more deeply nested and harder to read

Well, if you supply the expected message via `assertThrows`, then it would be printed even in the case the actual exception has a different class.

> It just seems to make the code more deeply nested and harder to read

Frankly speaking,

```java
    var exception = assertThrows(
      SignatureVerificationException.class,
      () -> {
        client.updateRoot();
        client.updateTimestamp();
      },
      "The timestamp was not signed so should have thown a SignatureVerificationException"
    );
    assertAll(
      () -> assertEquals(0, exception.getVerifiedSignatures()),
      () -> assertEquals(1, exception.getRequiredSignatures())
    );
```

looks pretty much the same to me as

```java
    try {
      client.updateRoot();
      client.updateTimestamp();
      fail("The timestamp was not signed so should have thown a SignatureVerificationException.");
    } catch (SignatureVerificationException e) {
      assertEquals(0, e.getVerifiedSignatures());
      assertEquals(1, e.getRequiredSignatures());
    }
```

Even though `try-catch` variation might look "easier to read", I believe, people rarely read tests. The test either just lives on its own, or you read **test failure**. In any case, that is, well, Java :)

Yet another possibility is to prefer `AssertJ` which provides better API and better error messages: https://www.baeldung.com/assertj-exception-assertion#1-using-assertthatthrownby

Truth yields better messages, yet it is not very compatible with JUnit 5.

---

At the same time, I don't understand  TUF tests and code, so if think `assertThrows` does not suit here, ignore it. However, I'm inclined  `assertThrows` is not that bad. However, I'm not going to veto `try-case`-based tests.

PS. The same in Kotlin would be:

```kotlin
import org.junit.jupiter.api.assertThrows // <-- out of the box JUnit

        val exception = assertThrows<SignatureVerificationException>(
            "The timestamp was not signed so should have thown a SignatureVerificationException"
        ) {
            client.updateRoot()
            client.updateTimestamp()
        }
        assertAll(
            { assertEquals(0, exception.verifiedSignatures) },
            { assertEquals(1, exception.requiredSignatures) },
        )
```

sigstore-java/src/main/java/dev/sigstore/tuf/FileSystemTufStore.java
```diff
@@ -41,18 +43,33 @@ public class FileSystemTufStore implements TufLocalStore {
     this.repoBaseDir = repoBaseDir;
   }
 
-  static TufLocalStore newFileSystemStore(Path repoBaseDir) {
+  public static TufLocalStore newFileSystemStore(Path repoBaseDir) {
     if (!repoBaseDir.toFile().isDirectory()) {
```
loosebazooka, line 47: You shouldn't need to convert to file here, `Files.isDirectory(repoBaseDir)`

sigstore-java/src/main/java/dev/sigstore/tuf/MetaNotFoundException.java
```diff
@@ -0,0 +1,24 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package dev.sigstore.tuf;
+
+/** Thrown when a metadata resources was unexpectedly missing. * */
```
loosebazooka, line 18: formatting `* */` -> `*/`

sigstore-java/src/main/java/dev/sigstore/tuf/TufClient.java
```diff
@@ -130,9 +129,7 @@ public void updateRoot()
     // 5.3.10) Check expiration timestamp in trusted root is higher than fixed update start time,
     // otherwise throw error.
     ZonedDateTime expires = trustedRoot.getSignedMeta().getExpiresAsDate();
-    if (expires.isBefore(updateStartTime)) {
-      throw new RootExpiredException(fetcher.getSource(), updateStartTime, expires);
-    }
+    throwIfExpired(expires);
```
loosebazooka, line 132: I don't really like this name, but whatever it's internal code.

sigstore-java/src/main/java/dev/sigstore/tuf/TufClient.java
```diff
@@ -206,17 +210,44 @@ void verifyDelegate(
     }
   }
 
-  public void updateTimestamp() {
-    // 1) download the timestamp.json bytes up to few 10s of K max.
+  public void updateTimestamp()
```
loosebazooka, line 213: @asraa or @znewman01 wanna take a quick look over this tuf workflow ?

sigstore-java/src/test/java/dev/sigstore/tuf/TufClientTest.java
```diff
@@ -141,8 +144,108 @@ public void testRootUpdate_metaFileTooBig()
...
     } catch (MetaFileExceedsMaxException e) {
+      // expected
+    }
+  }
+
+  @Test
+  public void testTimestampUpdate_throwMetaNotFoundException() throws IOException {
+    setupMirror("remote-timestamp-not-present");
+    var client = createTimeStaticTufClient(localStore);
+    assertThrows(
```
loosebazooka, line 157: I think you can short form this to (if you want):
assertThrows(MetaNotFoundException.class, client::updateTimestamp)

loosebazooka (commented):
some minor things, I want one TUF person to take a look at the logic, but otherwise lgtm

sigstore-java/src/main/java/dev/sigstore/tuf/TufClient.java
znewman01: in transit right now but will TAL in then next day or so

asraa: I'll take a look this evening!

asraa: func looks great to me!

loosebazooka (approved):

sigstore-java/src/main/java/dev/sigstore/tuf/TufClient.java
znewman01: i trust asra, listen to her

loosebazooka (approved):

PR #211, opened 2022-10-31
Add Error Prone verifications

vlsi:
The verification is enabled by default, and it can be skipped via
 `-PskipErrorprone` or adding `skipErrorprone=true` to `$HOME/.gradle/gradle.properties`

Error Prone produces actionable warnings, so I suggest we enable it by default.

See https://errorprone.info/bugpatterns

Some of the "experimental" patterns might be worth adding as well.

For instance,
* https://errorprone.info/bugpattern/FieldCanBeFinal
* https://errorprone.info/bugpattern/MethodCanBeStatic

`FieldCanBeFinal` yields several warnings in TUF code, so I did not enable it to avoid conflicts with @patflynn.


patflynn (dismissed):
nice!

patflynn (dismissed):

patflynn (approved):

PR #222, opened 2022-11-06
Optimize .idea/icon.png, add sandbox/.idea/* to gitignore

vlsi:
I've optimized the logo via https://tinypng.com/ and added `sandbox/.idea/*` to gitignore


loosebazooka:
think you need to commit with sign-off

loosebazooka (approved):

PR #237, opened 2022-12-09
Move to new key format

patflynn:
WIP towards #60 

While writing tests for https://github.com/sigstore/sigstore-java/tree/sync-snapshot I realised I would need to switch to generating my own TUF data. Specifically it wasn't possible to create a configuration with valid signatures and invalid hashes so I need control of the signing keys to produce that configuration.   That requirement  caused a snowball of changes that I'm pulling out as this PR.

A couple of fun facts:
- v5 of the root moves to PEM encoded keys which requires changing the verification logic slightly. Eventually I'll upgrade the bundle prod repo to v5 and remove the old key handling logic
- I had to fork https://github.com/cyberphone/json-canonicalization/tree/master/java/canonicalizer because the new TUF key format does not canonicalize according to the spec we were using


sigstore-java/src/main/java/dev/sigstore/json/canonicalizer/JsonCanonicalizer.java
```diff
@@ -0,0 +1,384 @@
+/*
+ * Copyright 2022 The Sigstore Authors.
```
loosebazooka, line 2: I believe you need to keep the original authors in and add a special case in our license handler. like: 
- https://github.com/sigstore/sigstore-java/blob/main/sigstore-java/build.gradle.kts#L92-L95
- https://github.com/sigstore/sigstore-java/blob/main/sigstore-java/src/main/java/dev/sigstore/encryption/certificates/transparency/CTConstants.java#L3

sigstore-java/src/main/java/dev/sigstore/tuf/Updater.java
```diff
@@ -139,6 +141,8 @@ Root updateRoot()
     throwIfExpired(expires);
     // 5.3.11) If the timestamp and / or snapshot keys have been rotated, then delete the trusted
     // timestamp and snapshot metadata files.
+    // TODO(patrick): have Asra look at this again. What if an old key is remove but there are no
```
loosebazooka, line 144: @asraa

sigstore-testkit/src/main/java/dev/sigstore/testkit/tuf/TestResources.java
```diff
@@ -25,8 +25,12 @@
 
 public class TestResources {
 
-  public static final Path UPDATER_TRUSTED_ROOT = Path.of(Resources.getResource("dev/sigstore/tuf/trusted-root.json").getPath());
-  public static final Path TUF_TEST_DATA_DIRECTORY = UPDATER_TRUSTED_ROOT.getParent();
+  public static final Path UPDATER_REAL_TRUSTED_ROOT = Path.of(Resources.getResource("dev/sigstore/tuf/real/trusted-root.json").getPath());
+  public static final Path UPDATER_SYNTHETIC_TRUSTED_ROOT = Path.of(Resources.getResource("dev/sigstore/tuf/synthetic/trusted-root.json").getPath());
+
+  // get the child folder and retrieve the parent to avoid resource loader finding the dev/sigstore/tuf folder in the classes/ dir.
```
loosebazooka, line 31: perhaps we need some better organization.

sigstore-java/src/main/java/dev/sigstore/tuf/Updater.java
```diff
@@ -154,7 +158,7 @@ private void throwIfExpired(ZonedDateTime expires) {
   }
 
   private boolean hasNewKeys(RootRole oldRole, RootRole newRole) {
-    return newRole.getKeyids().stream().allMatch(s -> oldRole.getKeyids().contains(s));
+    return !newRole.getKeyids().stream().allMatch(key -> oldRole.getKeyids().contains(key));
```
loosebazooka, line 161: was this a bug before?

loosebazooka (commented):
yeah seems fine to me

sigstore-java/src/main/java/dev/sigstore/tuf/Updater.java
asraa: I think you would fail verification of the snapshot/timestamp when updating those -- if there's no keys, then it can't hit the snapshot/timestamp signing threshold
i don't think that needs to be covered in this flow

sigstore-testkit/src/main/java/dev/sigstore/testkit/tuf/TestResources.java
loosebazooka: for another time though...

loosebazooka: can you just add a todo with an issue number linking to this in the issues?

sigstore-java/src/main/java/dev/sigstore/json/canonicalizer/JsonCanonicalizer.java
patflynn: thanks for the links! fixed.

sigstore-testkit/src/main/java/dev/sigstore/testkit/tuf/TestResources.java
patflynn: I've added a todo. I actually know how I can clean this up once I update the root in main.

sigstore-java/src/main/java/dev/sigstore/tuf/Updater.java
patflynn: Had to be. My updated tests caught it.

loosebazooka (approved):

sigstore-java/src/main/java/dev/sigstore/tuf/Updater.java
patflynn: Thanks for taking a look @asraa .  I think what wasn't clear to me from the spec is what constitutes 'rotating keys'.   I'm taking from your comment that I should leave as is, and 'rotating' is equivalent to adding one or more new keys regardless of whether old keys are kept or dropped.

loosebazooka (approved):

PR #245, opened 2022-12-12
Add conformance testing

### LINKED ISSUE #236: Add conformance testing
**Description**

Recently, @woodruffw and I have been working on a conformance testing suite for Sigstore clients. At the moment, there are a handful of tests but we'll be focusing on building out coverage in the coming months. The test suite is exposed via a GitHub Action which can be found [here](https://github.com/trailofbits/sigstore-conformance).

The way this works is that we've designed a [CLI protocol](https://github.com/trailofbits/sigstore-conformance/blob/main/docs/cli_protocol.md), which is essentially a set of flags that the client needs to support in order to be conformance tested. My understanding is that `sigstore-java` doesn't expose a CLI so, in order for this to work, there will need to be a barebones CLI app that supports the flags described in the linked doc.

If you're happy for `sigstore-java` to be included in this, I'd be happy to make a PR to get things going.

CC: @di

tetsuo-cpp:
Signed-off-by: Alex Cameron <asc@tetsuo.sh>

Closes #236

#### Summary

This PR adds a conformance testing workflow to `sigstore-java` as well as a CLI wrapper around the API in order for it to be tested.

#### Release Note

Added `sigstore-conformance` testing workflow.

#### Documentation

None.

CC: @woodruffw @di


tetsuo-cpp:
The conformance tests are passing on my fork's CI: https://github.com/trail-of-forks/sigstore-java/pull/1

.github/workflows/conformance.yml
```diff
@@ -0,0 +1,35 @@
...
+          java-version: 17
+          distribution: 'temurin'
+
+      - name: Build sigstore-java
+        uses: gradle/gradle-build-action@3fbe033aaae657f011f88f29be9e65ed26bd29ef # tag=v2.3.3
+        with:
+          arguments: clean :sigstore-conformance:build
+
+      - name: Unpack sigstore-java distribution
+        run: tar -xvf ${{ github.workspace }}/sigstore-conformance/build/distributions/sigstore-conformance-0.3.0.tar
```
tetsuo-cpp, line 31: Still need to figure out how I can do this in a way that doesn't hardcode the version number.

woodruffw: This isn't a fantastic solution, but maybe just a glob in a pinch?

```suggestion
        run: tar -xvf ${{ github.workspace }}/sigstore-conformance/build/distributions/sigstore-conformance-*.tar
```

(Assuming that there's only one archive matching that pattern at the end the build.)

woodruffw (approved):
I'm no Java expert so I can't review that for style/practice, but the CI parts LGTM!

loosebazooka:
can take a look wednesday if no one gets to it.

.github/workflows/conformance.yml
```diff
@@ -0,0 +1,35 @@
...
+jobs:
+  conformance:
+    permissions:
+      # Needed to access the workflow's OIDC identity.
+      id-token: write
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@d171c3b028d844f2bf14e9fdec0c58114451e4bf
+
+      - name: Set up JDK 17
```
loosebazooka, line 19: we also support java11, can we matrix this (I hope staging can keep up)?
https://github.com/sigstore/sigstore-java/blob/main/.github/workflows/ci.yaml#L31

sigstore-conformance/build.gradle.kts
```diff
@@ -0,0 +1,17 @@
...
+    id("java")
+    id("application")
+}
+
+repositories {
+    mavenCentral()
+}
+
+dependencies {
+    implementation(project(":sigstore-java"))
```
loosebazooka, line 11: curious if we should we also be testing against the latest release of sigstore-java? Since this is just at head.

loosebazooka (commented):
lgtm for now, just some minor things

.github/workflows/conformance.yml
woodruffw: Yep, having it be a matrix should be fine!

sigstore-conformance/build.gradle.kts
woodruffw: Hmm -- I don't think there's any problem with testing against the latest release, but I'm not sure it would add a ton of value either: the conformance suite mostly ensures that individual changes don't break the overall behavioral contract expected of Sigstore clients, so we'd expect a release that has already passed it once to continue passing it.

(We'd also end up de facto testing the latest release each time, by virtue of testing its `HEAD` commit right before the release is made. But we could make that explit/double it by also adding an event to test on the tag as well!)

.github/workflows/conformance.yml
tetsuo-cpp: [510a16e](https://github.com/sigstore/sigstore-java/pull/245/commits/510a16e6783fc39eef1e61aeb259bef02aa7c7ec)

sigstore-conformance/build.gradle.kts
tetsuo-cpp: I also think it makes more sense to test at head so you can detect non-conformant behaviour as soon as possible.

sigstore-conformance/build.gradle.kts
```diff
@@ -0,0 +1,17 @@
+plugins {
+    id("java")
+    id("application")
```
loosebazooka, line 3: id("com.diffplug.spotless") version "6.11.0"

sigstore-conformance/src/main/java/dev/sigstore/Main.java
```diff
@@ -0,0 +1,125 @@
+package dev.sigstore;
```
loosebazooka, line 1: This needs a copyright header. Preferably with enforcement from spotless (config added above).

`./gradlew spotlessApply` should apply spotless config to all files (and auto add license header)

sigstore-conformance/build.gradle.kts
```diff
@@ -0,0 +1,17 @@
...
+
+repositories {
+    mavenCentral()
+}
+
+dependencies {
+    implementation(project(":sigstore-java"))
+    implementation("com.google.guava:guava:31.1-jre")
+}
+
```
loosebazooka, line 14:
```
spotless {
    kotlinGradle {
        target("*.gradle.kts") // default target for kotlinGradle
        ktlint()
    }
    format("misc") {
        target("*.md", ".gitignore", "**/*.yaml")

        trimTrailingWhitespace()
        indentWithSpaces()
        endWithNewline()
    }
    java {
        googleJavaFormat("1.6")
        licenseHeaderFile("$rootDir/config/licenseHeader")
    }

}
```

loosebazooka (commented):
minor stuff that I missed last time

sigstore-conformance/build.gradle.kts
tetsuo-cpp: [8d90f65](https://github.com/sigstore/sigstore-java/pull/245/commits/8d90f653e915152454247099e0b51d6209d26c8c)

sigstore-conformance/build.gradle.kts
tetsuo-cpp: [8d90f65](https://github.com/sigstore/sigstore-java/pull/245/commits/8d90f653e915152454247099e0b51d6209d26c8c)

loosebazooka (dismissed):

loosebazooka:
I can't really expand the error that the framework is displaying as stdout.

tetsuo-cpp:
> I can't really expand the error that the framework is displaying as stdout. Looks like access to the oidc token in a forked PR thing?

Yeah that's right.

We've just cut v0.0.3 which solves this so I'll bump this PR to that version now.

sigstore-conformance/src/main/java/dev/sigstore/Main.java
```diff
@@ -0,0 +1,135 @@
...
+import static com.google.common.io.Files.asByteSource;
+import static dev.sigstore.encryption.certificates.Certificates.toPemString;
+
+import com.google.common.hash.Hashing;
+import dev.sigstore.oidc.client.GithubActionsOidcClient;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+
+public class Main {
```
vlsi, line 27: I imcline it would be better to select a less generic name than `dev.sigstore.Main` since ther might be several  tools for that.

How about `dev.sigstore.conformance.Main`?

.github/workflows/conformance.yml
```diff
@@ -0,0 +1,41 @@
...
+      - name: Set up JDK ${{ matrix.java-version }}
+        uses: actions/setup-java@v3
+        with:
+          java-version: ${{ matrix.java-version }}
+          distribution: 'temurin'
+
+      - name: Build sigstore-java
+        uses: gradle/gradle-build-action@3fbe033aaae657f011f88f29be9e65ed26bd29ef # tag=v2.3.3
+        with:
+          arguments: clean :sigstore-conformance:build
```
vlsi, line 34: There's no need to have `clean` here

tetsuo-cpp:
@loosebazooka 
Even after you approve this CI run, I believe it won't run since the `pull_request_target` trigger only works once its in the `main` branch. I have the conformance suite passing on my own fork: https://github.com/trail-of-forks/sigstore-java/pull/2

Since we've bumped to v0.0.3 and gained support for third party PRs, there are a few things to do on your side to get this working. You can refer to the instructions [here](https://github.com/trailofbits/sigstore-conformance#running-conformance-testing-in-pull-requests) but the general gist is:
- Create a label called "safe to test". This is used to mark trusted third party PRs and let the Action know that it is allowed to run since the conformance test suite will have access to the OIDC identity as discussed.
- Set the "fork pull request" permission to "require approval for all outside collaborators". I believe by default it is set to only require approval for first-time contributors.

sigstore-conformance/src/main/java/dev/sigstore/Main.java
tetsuo-cpp: Good point!

[59f0f96](https://github.com/sigstore/sigstore-java/pull/245/commits/59f0f96a2f9977d916757b1b43bf5f0c3d2f1380)

.github/workflows/conformance.yml
tetsuo-cpp: [f5f7d14](https://github.com/sigstore/sigstore-java/pull/245/commits/f5f7d140b60bb515f0313db0a1459a2f4e781afe)

vlsi:
I factored out Spotless configuration in https://github.com/sigstore/sigstore-java/pull/246, so all the `spotless {...}` boilerplate in this PR could be removed, and `id("java")` should be replaced with `id("build-logic.java")`

vlsi:
Frankly speaking, it sounds odd to pull Guava for "hashing a file", and using a hand-written argument parser at the same time :-)

I'm fine to merge the PR as is, however, I am inclined to either remove Guava dependency, or replace hand-written argument parser with something like https://rvesse.github.io/airline/, https://picocli.info/, or https://ajalt.github.io/clikt/ (and still remove Guava)

tetsuo-cpp:
> I factored out Spotless configuration in #246, so all the `spotless {...}` boilerplate in this PR could be removed, and `id("java")` should be replaced with `id("build-logic.java")`

Ok. Looks like that will be merged in soon so I'll refactor the config after its in `main`.

> Frankly speaking, it sounds odd to pull Guava for "hashing a file", and using a hand-written argument parser at the same time :-)
>
> I'm fine to merge the PR as is, however, I am inclined to either remove Guava dependency, or replace hand-written argument parser with something like https://rvesse.github.io/airline/, https://picocli.info/, or https://ajalt.github.io/clikt/ (and still remove Guava)

Haha point taken. Although I think Guava is already being pulled in as a dependency of one of the Google libraries that `sigstore-java` uses, as the same hashing code is already being used [here](https://github.com/sigstore/sigstore-java/blob/main/sigstore-java/src/main/java/dev/sigstore/KeylessSigner.java#L232). So sure it's adding a dependency to the conformance CLI, but it's not adding anything new to the project overall whereas an arg parsing library would be introducing a new dependency entirely.

The conformance suite is [designed](https://github.com/trailofbits/sigstore-conformance/blob/main/docs/cli_protocol.md#subcommands) to make this trivial to do by hand so I don't have to go around introducing dependencies to various clients. But if you and @loosebazooka are happy for me to introduce an arg parsing library,  I have no problem with refactoring this.

loosebazooka:
Uh I don't expect conformance to get very complicated so I'm fine with this. And this is purely testing so I'm also good importing whatever. I don't have strong opinions on dependencies for this module so I'm fine with it for now.

loosebazooka:
But yeah, if you have time to add picocli, go for it.

vlsi:
I see. If you consider this module only as a small shim for testing the validity if the actual java implementation, then it makes sense to refrain from custom third-party dependencies so we don't have extra renovate bot churn caused by upgrading Guava, picocli, etc

woodruffw:
Yep, that's the intended use/interaction with the conformance suite! We designed the entrypoint to be trivial and take arguments in a fixed order, so IMO a dependency for argument parsing is probably a little overkill 🙂

loosebazooka:
So I think I'm going to merge this for now. I'll remove guava and fix the build file in a followup. Thanks @tetsuo-cpp

loosebazooka (approved):

PR #247, opened 2022-12-16
switch to prod for rekor tests as staging is unstable

patflynn:


loosebazooka:
nice, can you change the gradle tests too?

loosebazooka (approved):

PR #251, opened 2022-12-17
Build updates to conformance module

loosebazooka:
- remove guava dep
- add build-logic.java (with error prone and spotless)

### DIFF
--- a/sigstore-conformance/src/main/java/dev/sigstore/conformance/Main.java
+++ b/sigstore-conformance/src/main/java/dev/sigstore/conformance/Main.java
@@ -15,16 +15,19 @@
  */
 package dev.sigstore.conformance;
 
-import static com.google.common.io.Files.asByteSource;
 import static dev.sigstore.encryption.certificates.Certificates.toPemString;
 
-import com.google.common.hash.Hashing;
 import dev.sigstore.KeylessSigner;
 import dev.sigstore.KeylessVerifier;
 import dev.sigstore.oidc.client.GithubActionsOidcClient;
+import java.io.IOException;
+import java.io.InputStream;
+import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
 
 public class Main {
   private static final String SIGN_COMMAND = "sign";
@@ -101,16 +104,20 @@ public class Main {
             .build();
     final var result = signer.signFile(args.artifact);
     Files.write(args.signature, result.getSignature());
-    final var pemBytes = toPemString(result.getCertPath()).getBytes();
+    final var pemBytes = toPemString(result.getCertPath()).getBytes(StandardCharsets.UTF_8);
     Files.write(args.certificate, pemBytes);
   }
 
   private static class VerifyArguments {
     public Path signature;
     public Path certificate;
-    public String certificateIdentity;
-    public String certificateOidcIssuer;
     public Path artifact;
+
+    @SuppressWarnings("unused") // remove when verifier actually verifies these
+    public String certificateIdentity;
+
+    @SuppressWarnings("unused") // remove when verifier actually verifies these
+    public String certificateOidcIssuer;
   }
 
   private static VerifyArguments parseVerifyArguments(Arguments args) {
@@ -129,9 +136,20 @@ public class Main {
 
   private static void executeVerify(VerifyArguments args) throws Exception {
     final var verifier = KeylessVerifier.builder().sigstorePublicDefaults().build();
-    final var artifactByteSource = asByteSource(args.artifact.toFile());
-    final byte[] artifactDigest = artifactByteSource.hash(Hashing.sha256()).asBytes();
+    final byte[] artifactDigest = sha256(args.artifact);
     verifier.verifyOnline(
         artifactDigest, Files.readAllBytes(args.certificate), Files.readAllBytes(args.signature));
+  }
+
+  private static byte[] sha256(Path path) throws IOException, NoSuchAlgorithmException {
+    MessageDigest digest = MessageDigest.getInstance("SHA-256");
+    try (InputStream in = Files.newInputStream(path)) {
+      byte[] buffer = new byte[1024];
+      int count;
+      while ((count = in.read(buffer)) > 0) {
+        digest.update(buffer, 0, count);
+      }
+    }
+    return digest.digest();
   }
 }
--- a/.github/workflows/conformance.yml
+++ b/.github/workflows/conformance.yml
@@ -20,10 +20,10 @@ jobs:
       id-token: write
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@d171c3b028d844f2bf14e9fdec0c58114451e4bf
+      - uses: actions/checkout@755da8c3cf115ac066823e79a1e1788f8940201b # tag=v3.2.0
 
       - name: Set up JDK ${{ matrix.java-version }}
-        uses: actions/setup-java@v3
+        uses: actions/setup-java@1df8dbefe2a8cbc99770194893dd902763bee34b # tag=v3.9.0
         with:
           java-version: ${{ matrix.java-version }}
           distribution: 'temurin'
@@ -36,6 +36,6 @@ jobs:
       - name: Unpack sigstore-java distribution
         run: tar -xvf ${{ github.workspace }}/sigstore-conformance/build/distributions/sigstore-conformance-*.tar --strip-components 1
 
-      - uses: trailofbits/sigstore-conformance@0748d63c53810e36cc3f4bbe4114301080f0d844
+      - uses: trailofbits/sigstore-conformance@0748d63c53810e36cc3f4bbe4114301080f0d844 # tag=v0.0.3
         with:
           entrypoint: ${{ github.workspace }}/bin/sigstore-conformance
--- a/sigstore-conformance/build.gradle.kts
+++ b/sigstore-conformance/build.gradle.kts
@@ -1,7 +1,6 @@
 plugins {
-    id("java")
+    id("build-logic.java")
     id("application")
-    id("com.diffplug.spotless") version "6.11.0"
 }
 
 repositories {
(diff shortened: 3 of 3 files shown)


tetsuo-cpp:
@loosebazooka Just a reminder, you need to create a "safe to test" label to trigger the conformance tests: https://github.com/sigstore/sigstore-java/pull/245#issuecomment-1354233387.

loosebazooka:
I'm a little confused @tetsuo-cpp , this is a first party PR. How are we expecting this to work? Put the label on every PR?

tetsuo-cpp:
> I'm a little confused @tetsuo-cpp , this is a first party PR. How are we expecting this to work? Put the label on every PR?

Yeah, at the moment it requires the label on any PR that you want to run conformance testing. We could try to do something smarter and check for the label only on forks, but I think it's better if it's consistent.

Does that seem reasonable to you?

vlsi:
What are the security implications of exposing oidc for untrusted scripts?

As far as I understand, machine id would include workflow name and branch name anyway, so the one who verifies should ignore such signatures anyway.

What is the actual reason for adding label to every PR?

sigstore-conformance/src/main/java/dev/sigstore/conformance/Main.java
```diff
@@ -129,9 +136,20 @@ private static VerifyArguments parseVerifyArguments(Arguments args) {
...
-    final byte[] artifactDigest = artifactByteSource.hash(Hashing.sha256()).asBytes();
+    final byte[] artifactDigest = sha256(args.artifact);
     verifier.verifyOnline(
         artifactDigest, Files.readAllBytes(args.certificate), Files.readAllBytes(args.signature));
   }
+
+  private static byte[] sha256(Path path) throws IOException, NoSuchAlgorithmException {
+    byte[] buffer = new byte[1024];
+    MessageDigest digest = MessageDigest.getInstance("SHA-256");
+    BufferedInputStream in = new BufferedInputStream(Files.newInputStream(path));
```
vlsi, line 147: Technically speaking, BufferedInputStream is not required since you use byte[1024] buffer anyway. It would be nice to have try-with-resources to close the stream.
However, those are tiny comments.

vlsi (approved):

sigstore-conformance/src/main/java/dev/sigstore/conformance/Main.java
loosebazooka: No I was lazy. I'll fix this.

loosebazooka:
@tetsuo-cpp I think the `safe-to-test` model is still a little prone to exploitation, for instance if I leave the label on a malicious submitter can just update the PR and execute bad code?

vlsi (approved):

tetsuo-cpp:
@vlsi

> What are the security implications of exposing oidc for untrusted scripts?
> 
> As far as I understand, machine id would include workflow name and branch name anyway, so the one who verifies should ignore such signatures anyway.

I don't think it includes the branch name, but it does contain a workflow ref which distinguishes between pull requests and releases: https://github.com/trailofbits/sigstore-conformance/issues/55#issuecomment-1343898291.

So you'd definitely be able to distinguish between an ID that is generated as part of a PR and one that would be used for a release workflow, which a user would use to verify a client distribution. Still, regardless of the workflow ref, the token is a secret by default and represents the repo (and its owners).

I can imagine a scenario where someone attempts to verify a `sigstore-java` release with an invocation pasted from somewhere with the correct release ID and verification fails with something like:

> found ID https://github.com/sigstore/sigstore-java/.github/workflows/conformance.yml@refs/pull/1/merge instead of https://github.com/sigstore/sigstore-java/.github/workflows/conformance.yml@release

A user might think "well it's a `sigstore-java` identity so that should be fine" and just paste the PR identity into the verify invocation. Maybe a bit farfetched, but within the realm of possibility.

To answer your question about "what are the security implications", there is no way for an attacker to assume the release identity. It's more a case of hygiene as well as the fact that an attacker can assume a very legitimate looking identity (but not the actual release one) which could cause some confusion.

> What is the actual reason for adding label to every PR?

It's a way for us to enforce these permissions from within the action. We recommend users of the action to set CI to require approval from all external contributors. However, there's no way to enforce this so we added the labelling mechanism as a safeguard.

I believe if you have your repo CI set to require approval for all external contributors, the labelling is redundant. But it feels like a bit of a footgun to make the action unsafe by default.

@loosebazooka 

> I think the safe-to-test model is still a little prone to exploitation, for instance if I leave the label on a malicious submitter can just update the PR and execute bad code?

At the moment, the `pull_request_target` is set to trigger on `labeled`. So it only runs when a label is added to the PR. In the case that a malicious submitter updates the PR to execute bad code, you'll have to remove and then add the label again to trigger the CI. It's unwieldy, but this shouldn't be a risk.

vlsi:
>A user might think "well it's a sigstore-java identity so that should be fine" and just paste the PR identity into the verify invocation. Maybe a bit farfetched, but within the realm of possibility.

I see what you mean, however, I inline that the ones who fail to validate sigstore properly are susceptible to many other dependency injection vulnerabilities.
For instance, what if they treat the mere presence of `.sigstore` signature as an indicator of a trusted signature?

So I incline that we should allow OIDC test for all PRs. We can reuse the same `pull_request_target + permissions:id-token:write` approach for testing GitHub OIDC in our regular test suite (currently it skips test if no OIDC is present).

tetsuo-cpp:
> So I incline that we should allow OIDC test for all PRs. We can reuse the same `pull_request_target + permissions:id-token:write` approach for testing GitHub OIDC in our regular test suite (currently it skips test if no OIDC is present).

I do see your point. When we discussed this, I think most of us were a bit uncomfortable with the idea of the identity being used by external collaborators even if there isn't a concrete idea of how this can be abused. I'm going to re-open https://github.com/trailofbits/sigstore-conformance/issues/55 to restart the discussion on this.

woodruffw:
FWIW, the other risk with `pull_request_target` is that you might run attacker supplied changes in the context of your own repository, rather than the attacker's fork. Meaning that, even if signing with the OIDC credential results in useless claims, they can exfiltrate workflow secrets or modify the `aud` claim during OIDC issuance to impersonate your repository to other services. 

Personally, I'd recommend maintaining the current "opt-in + label" flow -- it's unfortunately tedious, but it's more misuse resistant 🙂

vlsi:
> they can exfiltrate workflow secrets

They can not exfiltrate secrets in case you have `permissions: id-token: write`.
That explicitly drops all the other permissions, so they don't have access to secrets, they have no access to add commits, comments, etc.

woodruffw:
> That explicitly drops all the other permissions, so they don't have access to secrets, they have no access to add commits, comments, etc.

Do you have a source for that? That's great news if true, but I can't find it in GitHub's own docs.

(Even still, I think the potential for abuse justifies limiting `pull_request_target` -- there's still the `aud` impersonation problem, and it follows the principle of least use/privilege.)

loosebazooka:
hey y'all I'm going to merge this, but feel free to continue the discussion here or in a new issue.

vlsi:
> Do you have a source for that? That's great news if true, but I can't find it in GitHub's own docs

https://docs.github.com/en/actions/security-guides/automatic-token-authentication#modifying-the-permissions-for-the-github_token

See

> When the permissions key is used, all unspecified permissions are set to no access, with the exception of the metadata scope, which always gets read access.

woodruffw:
Thanks for the link! That's not quite the same (in my interpretation) as "drops all other permissions" -- the workflow in question might still have other permissions explicitly scoped, so `pull_request_target` is still potentially dangerous with untrusted code.

(There's also the "default set" problem -- if the org or enterprise hosting the repo has expansive `GITHUB_TOKEN` base permissions, then downscoping/dropping doesn't have much of an effect.)

vlsi:
> so pull_request_target is still potentially dangerous with untrusted code.

Of course, it is.
However, adding "label protection" does not make it tremendously more secure.

If a committer accidentally adds more permissions to the workflow on the main branch, it would be a separate vulnerability.
So I would recommend adding "never ever add more permissions" comment to the `.yaml` boilerplate template.

woodruffw:
> However, adding "label protection" does not make it tremendously more secure.

Agreed. The problem here is that `pull_request_target` is fundamentally insecure, but also currently the best way (in terms of convenience and reliably) to get an OIDC token for test suites that need to run on third-party PRs.

The goal here is to take an insecure thing and make it as misuse-resistant as possible. GitHub's own guidance for that is to use an explicit label, so IMO we should do that. Combined with requiring approval on every third-party PR, that represents the best currently known approach to securing these types of workflows.

vlsi:
I'm afraid humans have extremely poor track of keeping things secure.

I think there should be a "security validation" action that blocks attempts to increase the set of permissions, especially for workflows that have `pull_request_target`.

woodruffw:
Sure. But that's still an improvement over allowing anyone to run untrusted code unprompted, no? Unless I misunderstood, that's what you proposed in https://github.com/sigstore/sigstore-java/pull/251#issuecomment-1356016995.

I agree that there should be additional enforcement here, including a workflow that fails attempts to escalate the permission set for the conformance suite. But that's outside of the conformance suite itself.

vlsi:
What I suggest:
1. Just run conformance tests by default
2. Add GitHub Action that would ensure workflows that have `pull_request_target` do not accidentally allow permissions except `id-token`

So far I strongly believe that using the label "safe for test" provides a false sense of security, and it adds almost nothing even if used appropriately.

woodruffw:
I still don't think that's a good idea -- it turns a local configuration problem into a global one (orgs can have permissive default `GITHUB_TOKEN` settings), and it would allow someone to mint an OIDC token with any `aud` claim they like without interaction (which in turn could mean service impersonation). 

Having a separate enforcement action would mitigate the first problem, but not proactively -- someone could still run a `pull_request_target` action unprompted, access another service using OIDC impersonation, and then pivot that to a higher-permissioned token. Labels **do not** make that safe, but they **do** make it misuse-resistant, and that's a higher standard than the default.

TL;DR, my suggestion is:

1. Require maintainer approval on all third-party PRs that trigger `pull_request_target` workflows
2. Require *some* kind of label on `pull_request_target` workflows (I'm open to it being something other than "safe to test," if you think that language is misleading)
3. Add a GitHub Action that enforces the permission set, like you said (but note that it can't prevent default org and enterprise permissions, meaning that orgs themselves will have to comply with the workflow's requirements)

vlsi:
From my past labelling experience, I am sure I misclicked labels more than once.

For instance, suppose I want adding "need more info" label, however, I accidentally misclick and hit "safe for test".

At the same time, I find "safe for test" unclear (I think it is less important issue, and I think the label is not needed for the most common case). It does  not tell which safety is assumed, and it does not signal the consequence of accidental labelling.

loosebazooka:
Would it make sense to have separate workflows for 1st party and 3rd party PRs?

PR #257, opened 2022-12-19
clean up action references to use hashes

loosebazooka:


.github/workflows/ci.yaml
```diff
@@ -36,14 +36,14 @@ jobs:
     permissions:
       id-token: write
     steps:
-    - uses: actions/checkout@v3
+    - uses: actions/checkout@755da8c3cf115ac066823e79a1e1788f8940201b # tag=v3.2.0
```
loosebazooka, line 39: https://github.com/actions/checkout/commit/755da8c3cf115ac066823e79a1e1788f8940201b

.github/workflows/ci.yaml
```diff
@@ -36,14 +36,14 @@ jobs:
...
     - name: Set up JDK ${{ matrix.java-version }}
       uses: actions/setup-java@v3
       with:
         java-version: ${{ matrix.java-version }}
         distribution: 'temurin'
     - name: Setup Go environment
-      uses: actions/setup-go@v3.5.0
+      uses: actions/setup-go@6edd4406fa81c3da01a34fa6f6343087c207a568 # tag=v3.5.0
```
loosebazooka, line 46: https://github.com/actions/setup-go/commit/6edd4406fa81c3da01a34fa6f6343087c207a568

.github/workflows/tag-and-build-release.yaml
```diff
@@ -29,7 +29,7 @@ jobs:
       contents: write
     steps:
       - name: tag
-        uses: actions/github-script@v6
+        uses: actions/github-script@d556feaca394842dc55e4734bf3bb9f685482fa0 # tag=v6.3.3
```
loosebazooka, line 32: https://github.com/actions/github-script/commit/d556feaca394842dc55e4734bf3bb9f685482fa0

.github/workflows/tag-and-build-release.yaml
```diff
@@ -46,12 +46,12 @@ jobs:
...
     steps:
       - name: checkout tag
-        uses: actions/checkout@v3
+        uses: actions/checkout@755da8c3cf115ac066823e79a1e1788f8940201b # tag=v3.2.0
         with:
           ref: "refs/tags/v${{ github.event.inputs.release_version }}"
 
       - name: Set up JDK 11
-        uses: actions/setup-java@v3
+        uses: actions/setup-java@1df8dbefe2a8cbc99770194893dd902763bee34b # tag=v3.9.0
```
loosebazooka, line 54: https://github.com/actions/setup-java/commit/1df8dbefe2a8cbc99770194893dd902763bee34b

.github/workflows/tag-and-build-release.yaml
```diff
@@ -93,17 +94,17 @@ jobs:
       contents: write
     steps:
       - name: Download attestation
-        uses: actions/download-artifact@v3
+        uses: actions/download-artifact@9782bd6a9848b53b110e712e20e42d89988822b7 # tag=v3.0.1
```
loosebazooka, line 97: https://github.com/actions/download-artifact/commit/9782bd6a9848b53b110e712e20e42d89988822b7

.github/workflows/tag-and-build-release.yaml
```diff
@@ -93,17 +94,17 @@ jobs:
...
           path: ./release/
       - name: Download gradle release artifacts
-        uses: actions/download-artifact@v3
+        uses: actions/download-artifact@9782bd6a9848b53b110e712e20e42d89988822b7 # tag=v3.0.1
         with:
           name: project-release-artifacts
           path: ./release/
       - name: Create draft release
-        uses: softprops/action-gh-release@v1
+        uses: softprops/action-gh-release@de2c0eb89ae2a093876385947365aca7b0e5f844 # tag=v0.1.15
```
loosebazooka, line 107: https://github.com/softprops/action-gh-release/commit/de2c0eb89ae2a093876385947365aca7b0e5f844

vlsi:
What do you think of adding helpers:pinGitHubActionDigests to Renovate configuration instead?

See https://github.com/bazelbuild/rules_nodejs/blob/stable/renovate.json#L4

loosebazooka:
> What do you think of adding helpers:pinGitHubActionDigests to Renovate configuration instead?

seems fine, I like the informational comments though, can we just add that config and keep the `# tag=...` ? sometimes there are multiple action versions in concurrent development (ex: actions/checkout{v2,v3}) and I'd kind of like to know quickly which one we're using.

Will this also mess with the two exceptions for tagging we have in there (gradle wrapper validation, slsa?)

vlsi:
>I like the informational comments though,

As far as I understand, renovate would create such comments:
https://github.com/WhiteSourceE/renovate/blob/e8f68c4b63dfabb6081d7d1fa272f48e1379fd49/lib/manager/github-actions/readme.md

However, I suggest we just try enabling the configuration and check which PR would Renovate create.

> actions/checkout{v2,v3}

Technically speaking, they treat `actions/*` (or something like that) as an exception.
I guess the motivation is that "you implicitly agree that GitHub runner updates once every couple of weeks, then it should be just fine if `actions/checkout` updates as well.

> Will this also mess with the two exceptions for tagging we have in there (gradle wrapper validation, slsa)

Would you please clarify what exceptions you have in mind?

loosebazooka:
> If you want to automatically pin action digests add the helpers:pinGitHubActionDigests preset to the extends array:

I think the problem is here that it might covert some tags to hashes.

vlsi (approved):
Ok, got your point.
So we can't use Renovate for pinning since it does not seem to provide a configuration option to keep certain actions tag-based rather than sha-based

loosebazooka:
We can. If we add the tag comment.

PR #258, opened 2022-12-21
Use embedded-in-gradle-kotlin compiler version instead of hard-coded one

vlsi:
#### Summary

It unblocks upgrading Gradle as different Gradle embedd different Kotlin, so we should use the actually embedded one.

It would unblock https://github.com/sigstore/sigstore-java/pull/242

#### Release Note

NONE

#### Documentation

NONE

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -3,7 +3,7 @@ plugins {
     // It does not support participating in precompiled script plugins
     id("com.github.vlsi.stage-vote-release") version "1.84"
     // The Kotlin Gradle plugin was loaded multiple times in different subprojects, which is not supported and may break the build.
-    kotlin("jvm") version "1.5.31" apply false
+    `embedded-kotlin` apply false
 }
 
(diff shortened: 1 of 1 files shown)


build.gradle.kts
```diff
@@ -3,7 +3,7 @@ plugins {
     // It does not support participating in precompiled script plugins
     id("com.github.vlsi.stage-vote-release") version "1.84"
     // The Kotlin Gradle plugin was loaded multiple times in different subprojects, which is not supported and may break the build.
```
loosebazooka, line 5: Curious, if this comment still necessary

loosebazooka (approved):

build.gradle.kts
vlsi: Well, the comment kind of explains the reason we define `kotlin` here. The root project does not use Kotlin, so it might be tempting to remove unused bits. However, kotlin mention here is needed so it brings the plugin to the classpath once for all the subprojects (there's only one classpath anyway)

PR #264, opened 2022-12-26
Support Gradle Plugin testing with locally-built sigstore-java

vlsi:
#### Summary

The plugin runs with a separate build, so by default it uses mavenCentral only. However, we need to ensure the plugin is compatible with the main branch, so we publish sigstore-java to a temp folder, and use it as a repository in tests.

Ideally, we could parameterize over sigstore-java versions in the future, however, the currently published Central plugin does not support "building sigstore bundle", so a slightly better idea might be just requiring "matching" versions for now.

#### Release Note

NONE

#### Documentation

NONE

### DIFF
--- a/build-logic/publishing/src/main/kotlin/build-logic.kotlin-dsl-published-gradle-plugin.gradle.kts
+++ b/build-logic/publishing/src/main/kotlin/build-logic.kotlin-dsl-published-gradle-plugin.gradle.kts
@@ -7,3 +7,39 @@ plugins {
     id("build-logic.dokka-javadoc")
     id("build-logic.publish-to-central")
 }
+
+val sigstoreJavaRuntime by configurations.creating {
+    description = "declares dependencies that will be useful for testing purposes"
+    isCanBeConsumed = false
+    isCanBeResolved = false
+}
+
+val sigstoreJavaTestClasspath by configurations.creating {
+    description = "sigstore-java in local repository for testing purposes"
+    isCanBeConsumed = false
+    isCanBeResolved = true
+    extendsFrom(sigstoreJavaRuntime)
+    attributes {
+        attribute(Category.CATEGORY_ATTRIBUTE, objects.named("maven-repository"))
+        attribute(Bundling.BUNDLING_ATTRIBUTE, objects.named(Bundling.EXTERNAL))
+    }
+}
+
+tasks.test {
+    dependsOn(sigstoreJavaTestClasspath)
+    systemProperty("sigstore.test.current.version", version)
+    val projectDir = layout.projectDirectory.asFile
+    // This adds paths to the local repositories that contain currently-built sigstore-java
+    // It enables testing both "sigstore-java from Central" and "sigstore-java build locally" in the plugin tests
+    jvmArgumentProviders.add(
+        // Gradle does not support Provider for systemProperties yet, see https://github.com/gradle/gradle/issues/12247
+        CommandLineArgumentProvider {
+            listOf(
+                "-Dsigstore.test.local.maven.repo=" +
+                        sigstoreJavaTestClasspath.joinToString(File.pathSeparator) {
+                            it.toRelativeString(projectDir)
+                        }
+            )
+        }
+    )
+}
--- a/sigstore-testkit/src/main/kotlin/dev/sigstore/testkit/BaseGradleTest.kt
+++ b/sigstore-testkit/src/main/kotlin/dev/sigstore/testkit/BaseGradleTest.kt
@@ -26,6 +26,7 @@ import org.junit.jupiter.api.io.CleanupMode
 import org.junit.jupiter.api.io.TempDir
 import org.junit.jupiter.params.provider.Arguments
 import org.junit.jupiter.params.provider.Arguments.arguments
+import java.io.File
 import java.nio.file.Path
 
 open class BaseGradleTest {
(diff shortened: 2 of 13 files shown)


sigstore-testkit/src/main/kotlin/dev/sigstore/testkit/BaseGradleTest.kt
```diff
@@ -38,17 +39,44 @@ open class BaseGradleTest {
...
+            }
+        }
+
+        @JvmStatic
+        fun sigstoreJavaVersions(): Iterable<Arguments> {
+            return mutableListOf<Arguments>().apply {
+                add(arguments(SIGSTORE_JAVA_CURRENT_VERSION))
+                if (isCI) {
+                    add(arguments(TestedSigstoreJava.Default))
+                    add(arguments(TestedSigstoreJava.Version("0.2.0")))
```
loosebazooka, line 69: Is using "0.2.0" going to be hard to track?

sigstore-testkit/src/main/kotlin/dev/sigstore/testkit/TestedGradle.kt
```diff
@@ -0,0 +1,22 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+package dev.sigstore.testkit
+
```
loosebazooka, line 18: Can we add a comment here?

sigstore-testkit/src/main/kotlin/dev/sigstore/testkit/TestedSigstoreJava.kt
```diff
@@ -0,0 +1,24 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+package dev.sigstore.testkit
+
+sealed class TestedSigstoreJava {
```
loosebazooka, line 19: Can we add a comment to this class too?

build-logic/publishing/src/main/kotlin/build-logic.publish-to-local-repo.gradle.kts
```diff
@@ -0,0 +1,41 @@
...
+    builtBy(tasks.named("publishAllPublicationsToLocalRepository"))
+}
+
+val cleanLocalRepository by tasks.registering(Delete::class) {
+    description = "Clears local-maven-repo so timestamp-based snapshot artifacts do not consume space"
+    delete(localRepoDir)
+}
+
+tasks.withType<PublishToMavenRepository>()
+    .matching { it.name.contains("ToLocalRepository") }
```
loosebazooka, line 38: can we narrow the scope here?
perhaps `endsWith("...")` or `matches("...$")`

loosebazooka (commented):
I think this looks good, just some minor comments

build-logic/publishing/src/main/kotlin/build-logic.kotlin-dsl-published-gradle-plugin.gradle.kts
```diff
@@ -7,3 +7,46 @@ plugins {
     id("build-logic.dokka-javadoc")
     id("build-logic.publish-to-central")
 }
+
+val sigstoreJavaRuntime by configurations.creating {
```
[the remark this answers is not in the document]
vlsi: Sigstore-java does not need to reference "another sigstore-java varsion".

It is Gradle plugin that might need to reference multiple sigstore-java versions, so I added sigstoreJavaRuntime configuration to the shared logic of plugin scripts. sigstoreJavaRuntime is used to declare "which Maven repositories will be needed for the plugin". Then the plugin test code declares the repositories in the generated build.gradle, so the invoked script resolves the currently-built sigstore-java.

Does that resolve your question?

build-logic/publishing/src/main/kotlin/build-logic.kotlin-dsl-published-gradle-plugin.gradle.kts
```diff
@@ -7,3 +7,39 @@ plugins {
...
+    description = "declares dependencies that will be usefule for testing purposes"
+    isCanBeConsumed = false
```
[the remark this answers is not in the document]
vlsi: Java's getter has name of `isCanBeResolved` 🤷‍♂️ 
https://docs.gradle.org/current/javadoc/org/gradle/api/artifacts/Configuration.html#isCanBeResolved--

build-logic/publishing/src/main/kotlin/build-logic.kotlin-dsl-published-gradle-plugin.gradle.kts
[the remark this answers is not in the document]
loosebazooka: oh it's in "published-gradle-plugin" only. kk thanks for clarifying.

build-logic/publishing/src/main/kotlin/build-logic.publish-to-local-repo.gradle.kts
```diff
@@ -0,0 +1,41 @@
...
+        attribute(Bundling.BUNDLING_ATTRIBUTE, objects.named(Bundling.EXTERNAL))
+    }
+}
+
+val localRepoDir = layout.buildDirectory.dir("local-maven-repo")
+
+publishing {
+    repositories {
+        maven {
+            name = "local"
```
loosebazooka, line 22: I know we control the build, but would it make sense for this have a more unique name?

vlsi: Any suggestions?

loosebazooka: hah, yeah sorry I guess my comment was weakly actionable.

localProjectModules?

vlsi: The repo will include single module only :)
Each "module" creates its own repo under `build/...`, and publishes there.

It might be `tmp-maven-repo` would be better.

loosebazooka: Sgtm

sigstore-testkit/src/main/kotlin/dev/sigstore/testkit/BaseGradleTest.kt
vlsi: Any suggestions here?

loosebazooka: hrmm not really. maybe leave this like this for now?

loosebazooka (approved):

sigstore-testkit/src/main/kotlin/dev/sigstore/testkit/BaseGradleTest.kt
vlsi: The PR that adds bundle removes 0.2.0 from here.
I suggest we just keep it as is to see how it works before we move it to properties or something else.

loosebazooka:
Good to merge?

Issue #86, opened 2022-08-28
Consider byte[] for digests and "byte64" data


vlsi:
**Description**

The following would probably be better as `byte[]`:

```java
KeylessSigningResult {
  String getDigest();
```

```java
dev.sigstore.rekor.client.RekorEntry.InclusionProof {
List<String> getHashes();
```

loosebazooka:
This is direct from the json, but I can add derived methods that provide this in more palatable types (`getDigestBytes()` and `getHashesBytes()`) and maybe this will all just kind of go away if rekor adopts the grpc based communication

PR #271, opened 2023-01-03
Use byte[] for digest instead of hex String

loosebazooka:

needs #264
part of #86

### DIFF
--- a/sigstore-java/src/main/java/dev/sigstore/KeylessSigner.java
+++ b/sigstore-java/src/main/java/dev/sigstore/KeylessSigner.java
@@ -40,7 +40,6 @@ import java.security.spec.InvalidKeySpecException;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
-import org.bouncycastle.util.encoders.Hex;
 
 /** A full sigstore keyless signing flow. */
 public class KeylessSigner {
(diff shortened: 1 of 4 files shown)


sigstore-gradle/sigstore-gradle-sign-base-plugin/src/main/kotlin/dev/sigstore/sign/work/SignWorkAction.kt
```diff
@@ -65,10 +65,7 @@ abstract class SignWorkAction : WorkAction<SignWorkParameters> {
                 payloadHash = HashValue(
                     // See https://github.com/sigstore/sigstore-java/issues/85
                     algorithm = HashAlgorithm.sha256,
-                    // https://github.com/sigstore/sigstore-java/issues/86
-                    hash = result.digest.chunked(2)
-                        .map { it.toInt(16).toByte() }
-                        .toByteArray(),
+                    hash = result.digest,
```
vlsi, line 68: This would create merge conflict with bundle pr, so I would rather merge bundle the first

loosebazooka: Sure. Can wait on both

vlsi (approved):

Issue #292, opened 2023-01-14
Reuse signing certificates in gradle plugin


vlsi:
**Description**

Currently, gradle plugin signs individual files, so it asks for OIDC multiple times.
We need to reduce OIDC flows:
* once per module: it should be a low-hanging fruit in sign task
* reuse certificates across modules would be harder. Gradle shared build service might help there

PR #300, opened 2023-01-14
Reuse Fulcio certificates when the same signer is used multiple times

vlsi:
This reduces the number of OIDC flows

Fixes https://github.com/sigstore/sigstore-java/issues/292

#### Summary

See:
* https://github.com/sigstore/sigstore-java/issues/292

I add `dev.sigstore.KeylessSigner.Builder#minSigningCertificateLifetime(Duration)`, and make it `5 minutes` by default.
In other words, if the time allows, then the signer will reuse the certificate.

#### Release Note

KeylessSigner reuses Fulcio certificate if its lifetime exceeds a configurable threshold.

#### Documentation

NONE

### DIFF
--- a/sigstore-java/src/main/java/dev/sigstore/KeylessSigner.java
+++ b/sigstore-java/src/main/java/dev/sigstore/KeylessSigner.java
@@ -19,12 +19,16 @@ import com.google.api.client.util.Preconditions;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.hash.Hashing;
+import com.google.errorprone.annotations.CanIgnoreReturnValue;
+import com.google.errorprone.annotations.CheckReturnValue;
+import com.google.errorprone.annotations.concurrent.GuardedBy;
 import dev.sigstore.encryption.certificates.Certificates;
 import dev.sigstore.encryption.signers.Signer;
 import dev.sigstore.encryption.signers.Signers;
 import dev.sigstore.fulcio.client.*;
 import dev.sigstore.oidc.client.OidcClient;
 import dev.sigstore.oidc.client.OidcException;
+import dev.sigstore.oidc.client.OidcToken;
 import dev.sigstore.oidc.client.WebOidcClient;
 import dev.sigstore.rekor.client.*;
 import java.io.IOException;
@@ -37,18 +41,49 @@ import java.security.NoSuchAlgorithmException;
 import java.security.SignatureException;
 import java.security.cert.CertificateException;
 import java.security.spec.InvalidKeySpecException;
+import java.time.Duration;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.locks.ReentrantReadWriteLock;
+import org.checkerframework.checker.nullness.qual.Nullable;
 
-/** A full sigstore keyless signing flow. */
-public class KeylessSigner {
+/**
+ * A full sigstore keyless signing flow.
+ *
+ * <p>Note: the implementation is thread-safe assuming the clients (Fulcio, OIDC, Rekor) are
+ * thread-safe
+ */
+public class KeylessSigner implements AutoCloseable {
+  /**
+   * The instance of the {@link KeylessSigner} will try to reuse a previously acquired certificate
+   * if the expiration time on the certificate is more than {@code minSigningCertificateLifetime}
+   * time away. Otherwise, it will make a new request (OIDC, Fulcio) to obtain a new updated
+   * certificate to use for signing. This is a default value for the remaining lifetime of the
+   * signing certificate that is considered good enough.
+   */
+  public static final Duration DEFAULT_MIN_SIGNING_CERTIFICATE_LIFETIME = Duration.ofMinutes(5);
+
   private final FulcioClient fulcioClient;
   private final FulcioVerifier fulcioVerifier;
   private final RekorClient rekorClient;
   private final RekorVerifier rekorVerifier;
   private final OidcClient oidcClient;
   private final Signer signer;
+  private final Duration minSigningCertificateLifetime;
+
+  /** The code signing certificate from Fulcio. */
+  @GuardedBy("lock")
+  private @Nullable SigningCertificate signingCert;
+
+  /**
+   * Representation {@link #signingCert} in PEM bytes format. This is used to avoid serializing the
+   * certificate for each use.
+   */
+  @GuardedBy("lock")
+  private byte @Nullable [] signingCertPemBytes;
+
+  private final ReentrantReadWriteLock lock = new ReentrantReadWriteLock();
 
   private KeylessSigner(
       FulcioClient fulcioClient,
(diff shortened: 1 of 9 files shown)


sigstore-java/src/main/java/dev/sigstore/KeylessSigner.java
```diff
@@ -37,34 +40,49 @@
 import java.security.SignatureException;
 import java.security.cert.CertificateException;
 import java.security.spec.InvalidKeySpecException;
+import java.time.Duration;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
+import org.checkerframework.checker.nullness.qual.Nullable;
```
loosebazooka, line 47: javax.annotation is the one we've been using everywhere ?

sigstore-java/src/main/java/dev/sigstore/KeylessSigner.java
```diff
@@ -37,34 +40,49 @@
...
 
 /** A full sigstore keyless signing flow. */
-public class KeylessSigner {
+public class KeylessSigner implements AutoCloseable {
```
loosebazooka, line 50: Is autocloseable really necessary here? We're not holding any system resources or really anything of size. I assume GC at KeylessSigning cleanup is sufficient to release the memory used for the certs? Does this have to do with the gradle plugin signing service?

sigstore-java/src/main/java/dev/sigstore/KeylessSigner.java
```diff
@@ -76,40 +94,67 @@ public static class Builder {
...
 
+    @CanIgnoreReturnValue
     public Builder signer(Signer signer) {
       this.signer = signer;
       return this;
     }
 
+    /**
+     * The signer will request a signing certificate from Fulcio if the current certificate is going
+     * to expire within this duration.
```
loosebazooka, line 127: Can this be a bit more descriptive?
```
The instance of the KeylessSigner will try to reuse a previously acquired certificate if the expiration time on the certificate is more than "minSigningCertificateLifetime" time away. Otherwise, it will make a new request (oidc/fulcio) to obtain a new updated certificate to use for signing. Default "minSigningCertificateLifetime is {@value #SOME_STATIC_FIELD}". See https://docs.sigstore.dev/fulcio/overview/ for fulcio certificate validity.
```

sigstore-java/src/main/java/dev/sigstore/KeylessSigner.java
```diff
@@ -123,9 +168,11 @@ public Builder sigstorePublicDefaults()
           RekorVerifier.newRekorVerifier(VerificationMaterial.Production.rekorPublicKey()));
       oidcClient(WebOidcClient.builder().build());
       signer(Signers.newEcdsaSigner());
+      minSigningCertificateLifetime(Duration.ofMinutes(5));
```
loosebazooka, line 171: I think it's a bit confusing to set the default in the builder and here? Either make it a const in the class or remove this?

sigstore-java/src/main/java/dev/sigstore/KeylessSigner.java
```diff
@@ -141,20 +188,20 @@ public Builder sigstoreStagingDefaults()
...
     }
   }
 
   /**
    * Sign one or more artifact digests using the keyless signing workflow. The oidc/fulcio dance to
    * obtain a signing certificate will only occur once. The same ephemeral private key will be used
-   * to sign all artifacts. Errors may occur is the request is for an overwhelming number of
-   * artifactDigests as the certificate may expire -- this method does not current have the ability
-   * to obtain a new certificate if the one is use expires.
+   * to sign all artifacts. This method will verify and renew certificates as they expire.
```
loosebazooka, line 199: remove "verify and"?

sigstore-java/src/main/java/dev/sigstore/KeylessSigner.java
vlsi: `javax.annotation` is not `type` annotation. Currently, sigstore-java uses both checkerframework and jsr305.

The issue with jsr305 is the ideal annotation there is `@CheckNull` which is not very intuitive.
The meaning is that in jsr305, `@Nullable` means "might be null" while `@CheckNull` means "definitely can be null, so everybody should check for nulls".

sigstore-java/src/main/java/dev/sigstore/KeylessSigner.java
vlsi: Yeah, `AutoCloseable` is to release certificates (e.g. to avoid accidental reuse). It enables using keyless signer with Java's try-with-resources.

sigstore-java/src/main/java/dev/sigstore/KeylessSigner.java
loosebazooka: oh that's interesting, maybe `sigstorePublicDefaults` should be static on the KeylessSigner class rather than on the builder.

sigstore-java/src/main/java/dev/sigstore/KeylessSigner.java
loosebazooka: ah right, kk thanks for clarifying.

sigstore-java/src/main/java/dev/sigstore/KeylessSigner.java
```diff
@@ -141,20 +201,20 @@ public Builder sigstoreStagingDefaults()
...
    * obtain a signing certificate will only occur once. The same ephemeral private key will be used
-   * to sign all artifacts. Errors may occur is the request is for an overwhelming number of
-   * artifactDigests as the certificate may expire -- this method does not current have the ability
-   * to obtain a new certificate if the one is use expires.
+   * to sign all artifacts. This method will renew certificates as they expire.
    *
    * @param artifactDigests sha256 digests of the artifacts to sign.
    * @return a list of keyless singing results.
    */
+  @CheckReturnValue
```
loosebazooka, line 217: So this is now not thread safe? Should we make it threadsafe? I see that you use synchronized somewhere else?

Should we at least add that this is not threadsafe here to the javadoc?

vlsi: Nice catch. I've added `ReentrantReadWriteLock` to make certificate renewal thread-safe.
In the ideal case, we should probably add timeouts for `.lock()` methods and reuse, however, I believe it is fine to go with no timeout for `.lock()`.

sigstore-java/src/main/java/dev/sigstore/KeylessSigner.java
```diff
@@ -199,12 +285,49 @@ public List<KeylessSigningResult> sign(List<byte[]> artifactDigests)
...
+          FulcioVerificationException, OidcException {
+    lock.writeLock().lock();
+    try {
+      if (signingCert != null) {
+        @SuppressWarnings("JavaUtilDate")
+        long lifetimeLeft =
+            signingCert.getLeafCertificate().getNotAfter().getTime() - System.currentTimeMillis();
+        if (lifetimeLeft > minSigningCertificateLifetime.toMillis()) {
+          // The current certificate is fine, reuse it
+          return;
```
loosebazooka, line 300: I'm being pretty nitpicky here, but I believe this can be in a read lock? and then everything after can be in a write lock?

loosebazooka (approved):
lgtm, minor locking question

sigstore-java/src/main/java/dev/sigstore/KeylessSigner.java
vlsi: I have the same thoughts.

loosebazooka (approved):

vlsi:
Thanks for the review

PR #301, opened 2023-01-14
Avoid WorkValidationException when certain files do not exist for signing

vlsi:
#### Summary

Signing artifacts are generated for all the files (e.g. pom.xml, module.json, etc), however, if the task is disabled (e.g. someone disables Gradle Module Metadata), then module.json is not produced.

Gradle throws a validation error like "you are going to sign module.json, however, it is absent" Unfortunately, there's no easy way to include only those signing tasks that will be needed, so we generate signing tasks for all the files, and we just skip non-existing ones.

fixes https://github.com/sigstore/sigstore-java/issues/287

#### Release Note

NONE

#### Documentation

NONE

PR #302, opened 2023-01-14
Support Java toolchains in Gradle plugin

vlsi:
#### Summary

It enables selecting Java launcher for executing sigstore-java client.

#### Release Note

NONE

#### Documentation

NONE

### DIFF
--- a/sigstore-gradle/sigstore-gradle-sign-base-plugin/src/main/kotlin/dev/sigstore/sign/tasks/SigstoreSignFilesTask.kt
+++ b/sigstore-gradle/sigstore-gradle-sign-base-plugin/src/main/kotlin/dev/sigstore/sign/tasks/SigstoreSignFilesTask.kt
@@ -23,14 +23,19 @@ import dev.sigstore.sign.services.SigstoreSigningService
 import dev.sigstore.sign.work.SignWorkAction
 import org.gradle.api.Buildable
 import org.gradle.api.DefaultTask
+import org.gradle.api.JavaVersion
 import org.gradle.api.NamedDomainObjectContainer
 import org.gradle.api.file.*
 import org.gradle.api.model.ObjectFactory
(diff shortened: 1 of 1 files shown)


vlsi:
This will conflict with https://github.com/sigstore/sigstore-java/pull/301, so it will go after.

sigstore-gradle/sigstore-gradle-sign-base-plugin/src/main/kotlin/dev/sigstore/sign/tasks/SigstoreSignFilesTask.kt
```diff
@@ -159,10 +175,28 @@ abstract class SigstoreSignFilesTask : DefaultTask() {
     protected fun sign() {
         // Ensure task holds a lock, so no other signign tasks executes concurrently
```
loosebazooka, line 176: I think this comment maybe is unncessary now.

loosebazooka (approved):

sigstore-gradle/sigstore-gradle-sign-base-plugin/src/main/kotlin/dev/sigstore/sign/tasks/SigstoreSignFilesTask.kt
vlsi: It is a comment to signingservice.get
Serialization is needed so Gradle does not spawn multiple worker processes as each process would perform its own OIDC dance

loosebazooka:
do you want to merge this?

vlsi:
>Would the gradle sigstore plugin fail to load before we hit this? since it's built with java11? oh jk, it's pure kotlin?

The toolchain support would enable users to select different JDK.
For instance, if someone wants to launch Gradle builds with Java 8, they will need a way to configure sigstore task to run with Java 11 or 17.

PR #305, opened 2023-01-14
Update jsonscheme2dataclass plugin to 6.0.0

loosebazooka:

### DIFF
--- a/sigstore-java/build.gradle.kts
+++ b/sigstore-java/build.gradle.kts
@@ -3,7 +3,7 @@ import com.google.protobuf.gradle.id
 plugins {
     id("build-logic.java-published-library")
     id("build-logic.test-junit5")
-    id("org.jsonschema2dataclass") version "5.0.0"
+    id("org.jsonschema2dataclass") version "6.0.0"
     id("com.google.protobuf") version "0.9.2"
 }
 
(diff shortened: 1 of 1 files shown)


vlsi (approved):

vlsi:
For reference, the failure is caused by empty `@param data` javadocs generated on constructors as follows: https://github.com/joelittlejohn/jsonschema2pojo/issues/1463

```java
    /**
     * 
     * @param data
     * @param signature
     */
    public HashedRekord(Signature signature, Data data) {
        super();
        this.signature = signature;
        this.data = data;
    }
```

loosebazooka:
I think I'm going to ignore the javadoc generation on this subpath for now.

loosebazooka:
javadoc is really annoying to configure. Might sit on this for a while. Another option is change the package of the generated model from dev.sigstore.rekor -> dev.sigstore.rekor.model and exclude just that package path (easy)

vlsi:
@loosebazooka , another idea: add regexp-based processing to kill ill javadocs or replace them with `@param data data`.

WDYT?

vlsi:
Using a separate package for generated code looks good to me.

loosebazooka:
> @loosebazooka , another idea: add regexp-based processing to kill ill javadocs or replace them with @param data data.

Hah, I think I would rather not do this :sweat_smile:

PR #307, opened 2023-01-16
Adds snapshot and target updates as well as downloading of targets.

patflynn:
More progress towards #60.

This PR:

- updates the snapshot resource
- updates the targets resource
- downloads targets to a targets cache that defaults to {localTufRepo}/targets

There is a noteworthy todo. In the rather long span of time between when I started this work and today, sigstore-tuf has moved to using [consistent snapshot](https://theupdateframework.github.io/specification/latest/#consistent-snapshots) for all its resources.  It's not a major change but I'd rather not further delay/complicate this PR and fix it in a follow PR.

@asraa I'll tag you in a few areas where I was hoping for your input. :pray:


.editorconfig
```diff
@@ -22,3 +22,6 @@ ij_continuation_indent_size = 4
 # Doc: https://youtrack.jetbrains.com/issue/IDEA-170643#focus=streamItem-27-3708697.0-0
 # $ means "static"
 ij_java_imports_layout = $*,|,*
+
+[*.json]
+insert_final_newline = false
```
vlsi, line 27: Can we rather add a final newline in the committed files?
Otherwise Git (and GitHub UI) emit warnings on the missing newlines :-/

vlsi:
Did you intend to commit `.fleet/settings.json` and `bla` files?

.editorconfig
patflynn: Sadly no. This is the way that TUF outputs them and the signing and hash checks depend on them. :( Is there a way  to scope down the editor config to a path?

patflynn:
@vlsi Nope! removing.

.editorconfig
vlsi: You can put file path into the square brackets: https://editorconfig.org/#file-format-details

For instance, you can put `[test/resources/dev/sigstore/tuf/**/*.json]`, or you can add a new `.editorconfig` file (e.g. inside `test/resources/dev/sigstore/tuf` directory).

vlsi: In any case, please add a comment above that `insert_final_newline` so the others see the reason.

patflynn: done

sigstore-java/src/main/java/dev/sigstore/tuf/Updater.java
```diff
@@ -78,24 +76,34 @@ public static Builder builder() {
...
           SignatureVerificationException {
     // 5.3.1) record the time at start and use for expiration checks consistently throughout the
     // update.
     updateStartTime = ZonedDateTime.now(clock);
 
     // 5.3.2) load the trust metadata file (root.json), get version of root.json and the role
     // signature threshold value
-
-    Root trustedRoot = GSON.get().fromJson(Files.readString(trustedRootPath), Root.class);
+    // TODO review from asraa@ do we need to verify the local root against our trusted root from the
```
patflynn, line 98: @asraa FYI

sigstore-java/src/main/java/dev/sigstore/tuf/Updater.java
```diff
@@ -225,82 +234,190 @@ void verifyDelegate(
...
   }
 
-  Snapshot updateSnapshot(Root root, Timestamp timestamp) {
+  Snapshot updateSnapshot(Root root, Timestamp timestamp)
+      throws IOException, FileNotFoundException, InvalidHashesException,
+          SignatureVerificationException, NoSuchAlgorithmException, InvalidKeySpecException,
+          InvalidKeyException {
     // 1) download the snapshot.json bytes up to few 10s of K max.
-
+    // TODO:asraa@ to review whether we need to be using the snapshot size from the timestamp file
```
patflynn, line 278: @asraa

sigstore-java/src/main/java/dev/sigstore/tuf/Updater.java
```diff
@@ -225,82 +234,190 @@ void verifyDelegate(
...
-    // 9) Download target up to length specified in bytes. verify against hash.
-
-    // Done!!
-    return null;
+  void downloadTargets(Targets targets)
+      throws IOException, TargetMetadataMissingException, FileNotFoundException {
+    // Skip #7 and go straight to downloading targets. It looks like delegations were removed from
+    // sigstore TUF data.
+    // {@see https://github.com/sigstore/sigstore/issues/562}
+    // TODO(patrick@chainguard.dev): get review from Asra on whether we should be processing
```
patflynn, line 398: @asraa

sigstore-java/src/main/java/dev/sigstore/tuf/Updater.java
```diff
@@ -225,82 +234,190 @@ void verifyDelegate(
...
+      // to parse. Leaving
+      // this code in in-case we eventually allow it in de-serialization.
+      if (entry.getValue() == null) {
+        throw new TargetMetadataMissingException(targetName);
+      }
+      TargetMeta.TargetData targetData = entry.getValue();
+      // 9) Download target up to length specified in bytes. verify against hash.
+      // TODO(patrick@chainguard.dev): Update this code to use consistent snapshots.
+      // e.g. "{targetData.getHashes().getSha512()}.{targetName}"
+      var targetBytes = fetcher.fetchResource("targets/" + targetName, targetData.getLength());
```
vlsi, line 414: Should targetName be url-encoded?
Url concatenation sounds unsafe

sigstore-java/src/test/java/dev/sigstore/tuf/FileSystemTufStoreTest.java
```diff
@@ -45,7 +45,7 @@ void setTrustedRoot_noPrevious(@TempDir Path repoBase) throws IOException {
     TufLocalStore tufLocalStore = FileSystemTufStore.newFileSystemStore(repoBase);
     assertFalse(repoBase.resolve("root.json").toFile().exists());
     tufLocalStore.storeTrustedRoot(TestResources.loadRoot(TestResources.UPDATER_REAL_TRUSTED_ROOT));
-    assertEquals(1, repoBase.toFile().list().length);
+    assertEquals(2, repoBase.toFile().list().length); // root.json plus the targets dir.
```
vlsi, line 48: Could you move code comment to the assertion message?

patflynn: done.

sigstore-java/src/main/java/dev/sigstore/tuf/Updater.java
patflynn: :thinking: the fetchResource implementation passes the path to GenericUrl which will also concatenates the path simply as a string to the mirror context.  I presumed that GenericUrl handles the url-encoding and looking into the code it appears to do so (despite the parameter name).  The input is trusted, what are your exact thoughts on the dangers and what do you have in mind as an alternative?  Building every component of the path from URLs?

vlsi: I might need looking closer, however, building urls via string concatenation is always suspicious

sigstore-java/src/main/java/dev/sigstore/tuf/Updater.java
asraa: Does "from the client" mean the embedded trust root? If so, no - because it may not be possible and the local root *should* be trusted (that being said, maybe there could be a fallback mechanism on the client root in case `updateRoots()` using the local trusted root doesn't pass)

Why it may not be possible: if the client embedded root is 1.root.json, and the local updated to version 3, then you've lost the 2.root.json that you needed in order to verify the chain. It would still exist on the remote and you could -- but then you're better off ALWAYS starting from the embedded root and always refreshing your cache. I'd say assume your cache is trusted

sigstore-java/src/main/java/dev/sigstore/tuf/Updater.java
asraa: Yes! You can use the number of bytes int the timestamp file.

sigstore-java/src/main/java/dev/sigstore/tuf/Updater.java
asraa: Yes, I'd skip delegations. Delegations have a tricky model (where higher targets files take precedence) and so we deferred these. We'll likely reintroduce this, but very slowly (starting with an optional one only relevant to npm, and they will probably set the starter example for their TUF client)

asraa (commented):
nice!!

sigstore-java/src/main/java/dev/sigstore/tuf/Updater.java
patflynn: Thanks for the detailed response! removing comment and leaving as is.

sigstore-java/src/main/java/dev/sigstore/tuf/Updater.java
patflynn: updated to use max bytes to limit snapshot and targets download.

sigstore-java/src/main/java/dev/sigstore/tuf/Updater.java
patflynn: ack.

patflynn:
@vlsi @asraa any further comments? @loosebazooka is out for the week. mentioned he's happy if @vlsi approves.

sigstore-java/src/main/java/dev/sigstore/tuf/Updater.java
asraa: Agree that's the best case for now.

asraa (dismissed):
lgtm! there's an outstanding TODO around enabling target fetching by consistent snapshots, so I recommend that since prod now uses that.
but youre aware of that by the docs  :)

vlsi (approved):
LGTM
There are style comments, however, they are not vital.
Special bonus goes for documenting the way the test data was generated

sigstore-java/src/main/java/dev/sigstore/tuf/InvalidHashesException.java
```diff
@@ -0,0 +1,59 @@
...
+ * limitations under the License.
+ */
+package dev.sigstore.tuf;
+
+/** Thrown when a hash check fails for a given resource. */
+public class InvalidHashesException extends TufException {
+  static class InvalidHash {
+    String algorithm;
+    String expectedHash;
+    String computedHash;
```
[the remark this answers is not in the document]
patflynn: done.

sigstore-java/src/main/java/dev/sigstore/tuf/InvalidHashesException.java
```diff
@@ -0,0 +1,59 @@
...
+            resourceName, invalidHashesToString(invalidHashes)));
+  }
+
+  private static String invalidHashesToString(InvalidHash... invalidHashes) {
+    StringBuilder stringBuilder = new StringBuilder();
+    for (InvalidHash invalidHash : invalidHashes) {
+      stringBuilder.append(invalidHash);
+      stringBuilder.append('\n');
+    }
+    return stringBuilder.toString();
```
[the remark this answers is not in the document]
patflynn: doh. I'm out of practice.

sigstore-java/src/main/java/dev/sigstore/tuf/Updater.java
```diff
@@ -78,24 +76,32 @@ public static Builder builder() {
...
     // signature threshold value
-
-    Root trustedRoot = GSON.get().fromJson(Files.readString(trustedRootPath), Root.class);
+    Optional<Root> localRoot = localStore.loadTrustedRoot();
+    Root trustedRoot;
+    if (localRoot.isPresent()) {
+      trustedRoot = localRoot.get();
+    } else {
+      trustedRoot = GSON.get().fromJson(Files.readString(trustedRootPath), Root.class);
+    }
```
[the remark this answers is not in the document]
patflynn: I like that a lot! If it looks like I haven't coded since Java 7, you'd be right. :)  

Slight issue though, when I move to that approach the IOException from Files.readString becomes a problem for the lambda. Any tricks for cleanly passing that through up the stack?

sigstore-java/src/main/java/dev/sigstore/tuf/Updater.java
```diff
@@ -225,82 +232,194 @@ void verifyDelegate(
...
 
-    // 3) check that version of new timestamp.json is higher or equal than current, else fail.
-    //     3.2) check that timestamp.snapshot.version <= timestamp.version or fail
+    // 3) If the new timestamp file has a lesser version than our current trusted timestamp file
+    // report a rollback attack.  If it is equal abort the update as there should be no changes. If
+    // it is higher than continue update.
     Optional<Timestamp> localTimestampMaybe = localStore.loadTimestamp();
     if (localTimestampMaybe.isPresent()) {
       Timestamp localTimestamp = localTimestampMaybe.get();
       if (localTimestampMaybe.get().getSignedMeta().getVersion()
```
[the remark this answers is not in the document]
patflynn: good catch. they were redundant.

sigstore-java/src/main/java/dev/sigstore/tuf/Updater.java
```diff
@@ -225,82 +232,194 @@ void verifyDelegate(
...
+  // this method feels very wrong. I would not show it to a friend.
+  @VisibleForTesting
+  static void verifyHashes(String name, byte[] data, Hashes hashes) throws InvalidHashesException {
+    List<InvalidHashesException.InvalidHash> badHashes = new ArrayList<>(2);
+    String expectedSha512 = hashes.getSha512();
+    String expectedSha256 = hashes.getSha256();
+    if (expectedSha256 == null && expectedSha512 == null) {
+      throw new IllegalArgumentException(
+          String.format(
+              "Specified hashes for %s is missing either a sha256 or sha512 expected hash value",
```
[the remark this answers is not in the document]
patflynn: fixed.

vlsi (approved):

vlsi (approved):

vlsi (approved):

sigstore-java/src/main/java/dev/sigstore/tuf/Updater.java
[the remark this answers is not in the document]
vlsi: I missed there's a checked exception

PR #309, opened 2023-01-18
Only run if specific label is applied

loosebazooka:


vlsi (approved):

loosebazooka:
hrmm I'm not sure if that worked or not

loosebazooka:
hey it does work! but it's a bit nasty if you use a lot of other labels.

PR #311, opened 2023-01-18
Revert "Only run if specific label is applied"

loosebazooka:
Reverts sigstore/sigstore-java#309

woops, turns out this is baked into conformance suite itself: https://github.com/trailofbits/sigstore-conformance/blob/b76987fe1dce0e29476ace3bbaf20719f1a646cb/action.yml#L18-L23

### DIFF
--- a/.github/workflows/conformance.yml
+++ b/.github/workflows/conformance.yml
@@ -10,7 +10,6 @@ on:
 
 jobs:
   conformance:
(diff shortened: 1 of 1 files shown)


loosebazooka:
also I definitely though it was `safe-to-test` instead of `safe to test` and now everything is blocked.

vlsi (approved):

PR #326, opened 2023-01-29
Fetch Fulcio and Rekor certificates from TUF store

vlsi:
#### Summary

The idea is:
1. replace current `VerificationMaterial` (see https://github.com/sigstore/sigstore-java/blob/5539b01a017b17c5243a080eb00c528b082a0074/sigstore-java/src/main/java/dev/sigstore/VerificationMaterial.java ) class with a TUF repository
2. and make TUF repository auto-updateable (e.g. if the root rotates during we sign)

It builds upon https://github.com/sigstore/sigstore-java/pull/325, so check the second commit in the PR.

It looks like we need to deprecate:
1. `dev.sigstore.VerificationMaterial`
2. `dev.sigstore.fulcio.client.FulcioVerifier#newFulcioVerifier(byte[] fulcioRoot, List<byte[]> ctfePublicKeys)`. I suggest replacing it with `newFulcioVerifier(TufBasedVerificationMaterial)`
3. `dev.sigstore.rekor.client.RekorVerifier#newRekorVerifier(byte[] rekorPublicKey)` -> `newRekorVerifier(TufBasedVerificationMaterial)`

#### Release Note

Support Fulcio, Rekor, CTFE certificate rotation via TUF.

#### Documentation

NONE

### DIFF
--- a/sigstore-java/src/main/java/dev/sigstore/tuf/Updater.java
+++ b/sigstore-java/src/main/java/dev/sigstore/tuf/Updater.java
@@ -58,18 +58,15 @@ public class Updater {
   private Verifiers.Supplier verifiers;
   private MetaFetcher fetcher;
   private ZonedDateTime updateStartTime;
-  private Path trustedRootPath;
   private MutableTufStore localStore;
 
   Updater(
       Clock clock,
       Verifiers.Supplier verifiers,
       MetaFetcher fetcher,
-      Path trustedRootPath,
       MutableTufStore localStore) {
     this.clock = clock;
     this.verifiers = verifiers;
-    this.trustedRootPath = trustedRootPath;
     this.localStore = localStore;
     this.fetcher = fetcher;
   }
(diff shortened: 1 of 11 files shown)


sigstore-java/src/main/java/dev/sigstore/tuf/Updater.java
```diff
@@ -100,13 +97,7 @@ Root updateRoot()
 
     // 5.3.2) load the trust metadata file (root.json), get version of root.json and the role
     // signature threshold value
-    Optional<Root> localRoot = localStore.loadTrustedRoot();
-    Root trustedRoot;
-    if (localRoot.isPresent()) {
-      trustedRoot = localRoot.get();
-    } else {
-      trustedRoot = GSON.get().fromJson(Files.readString(trustedRootPath), Root.class);
```
vlsi, line 108: It sounds super-strange to have a TUF store without a root. I believe `loadTrustedRoot` should return non-`Optional` value.

@patflynn ?

patflynn: The local TUF store can be empty until a successful update has occurred. A trusted root needs to be passed to the updater along with the remote url, but that trusted root is bundled with the client or provided by the user and doesn't typically get copied into the store.

loosebazooka:
I believe this is now handled

PR #328, opened 2023-02-01
Add backoff on 429

### LINKED ISSUE #327: Handle 429 Too Many Requests from Rekor
**Description**

Apparently, the team behind google-http-java-client does not want to handle 429 by default: https://github.com/googleapis/google-http-java-client/issues?q=429

Stacktrace:

```
Caused by: com.google.api.client.http.HttpResponseException: 429 Too Many Requests
    POST https://rekor.sigstore.dev/api/v1/log/entries
    <html>
    <head><title>429 Too Many Requests</title></head>
    <body>
    <center><h1>429 Too Many Requests</h1></center>
    <hr><center>nginx</center>
    </body>
    </html>
    
    	at com.google.api.client.http.HttpResponseException$Builder.build(HttpResponseException.java:293)
    	at com.google.api.client.http.HttpRequest.execute(HttpRequest.java:1118)
    	at dev.sigstore.rekor.client.RekorClient.putEntry(RekorClient.java:88)
    	at dev.sigstore.KeylessSigner.sign(KeylessSigner.java:274)
    	at dev.sigstore.KeylessSigner.signFiles(KeylessSigner.java:365)
    	at dev.sigstore.KeylessSigner.signFile(KeylessSigner.java:384)
    	at dev.sigstore.sign.work.SignWorkAction.execute(SignWorkAction.kt:57)
    	at org.gradle.workers.internal.DefaultWorkerServer.execute(DefaultWorkerServer.java:63)
    	at org.gradle.workers.internal.AbstractClassLoaderWorker$1.create(AbstractClassLoaderWorker.java:49)
    	at org.gradle.workers.internal.AbstractClassLoaderWorker$1.create(AbstractClassLoaderWorker.java:43)
    	at org.gradle.internal.classloader.ClassLoaderUtils.executeInClassloader(ClassLoaderUtils.java:97)
    	at org.gradle.workers.internal.AbstractClassLoaderWorker.executeInClassLoader(AbstractClassLoaderWorker.java:43)
    	at org.gradle.workers.internal.IsolatedClassloaderWorker.run(IsolatedClassloaderWorker.java:49)
    	at org.gradle.workers.internal.IsolatedClassloaderWorker.run(IsolatedClassloaderWorker.java:30)
    	at org.gradle.workers.internal.WorkerDaemonServer.run(WorkerDaemonServer.java:85)
    	at org.gradle.workers.internal.WorkerDaemonServer.run(WorkerDaemonServer.java:55)
    	at org.gradle.process.internal.worker.request.WorkerAction$1.call(WorkerAction.java:138)
    	at org.gradle.process.internal.worker.child.WorkerLogEventListener.withWorkerLoggingProtocol(WorkerLogEventListener.java:41)
    	at org.gradle.process.internal.worker.request.WorkerAction.run(WorkerAction.java:135)
    	at jdk.internal.reflect.GeneratedMethodAccessor45.invoke(Unknown Source)
    	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
    	at org.gradle.internal.dispatch.ReflectionDispatch.dispatch(ReflectionDispatch.java:36)
    	at org.gradle.internal.dispatch.ReflectionDispatch.dispatch(ReflectionDispatch.java:24)
    	at org.gradle.internal.remote.internal.hub.MessageHubBackedObjectConnection$DispatchWrapper.dispatch(MessageHubBackedObjectConnection.java:182)
    	at org.gradle.internal.remote.internal.hub.MessageHubBackedObjectConnection$DispatchWrapper.dispatch(MessageHubBackedObjectConnection.java:164)
    	at org.gradle.internal.remote.internal.hub.MessageHub$Handler.run(MessageHub.java:414)
    	at org.gradle.internal.concurrent.ExecutorPolicy$CatchAndRecordFailures.onExecute(ExecutorPolicy.java:64)
    	at org.gradle.internal.concurrent.ManagedExecutorImpl$1.run(ManagedExecutorImpl.java:48)
    	at org.gradle.internal.concurrent.ThreadFactoryImpl$ManagedThreadRunnable.run(ThreadFactoryImpl.java:61)
```

loosebazooka:
hopefully fixes #327

### DIFF
--- a/sigstore-java/src/main/java/dev/sigstore/http/UnsuccessfulResponseHandler.java
+++ b/sigstore-java/src/main/java/dev/sigstore/http/UnsuccessfulResponseHandler.java
@@ -1,0 +1,124 @@
+/*
+ * Copyright 2023 The Sigstore Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package dev.sigstore.http;
+
+import com.google.api.client.http.HttpRequest;
+import com.google.api.client.http.HttpResponse;
+import com.google.api.client.http.HttpUnsuccessfulResponseHandler;
+import com.google.api.client.util.BackOff;
+import com.google.api.client.util.BackOffUtils;
+import com.google.api.client.util.ExponentialBackOff;
+import com.google.api.client.util.Sleeper;
+import com.google.common.annotations.VisibleForTesting;
+import java.io.IOException;
+import java.time.Instant;
+import java.util.logging.Logger;
+import org.apache.http.client.utils.DateUtils;
+
+/**
+ * Handler for 429 and standard server errors 5XX. Objects of this class are single use. A new
+ * instance should be created for each request (but this is handled in {@link
+ * HttpClients#newRequestFactory(HttpParams)}).
+ *
+ * <p>This implementation is a bit weird, but hopefully we can just move to grpc for rekor when
+ * ready.
+ */
+public class UnsuccessfulResponseHandler implements HttpUnsuccessfulResponseHandler {
+  // handler may return before 10 retries if exponentialBackoff hits its limit first
+  private static final int MAX_RETRIES = 10;
+  private static final Logger log = Logger.getLogger(UnsuccessfulResponseHandler.class.getName());
+
+  private final Sleeper sleeper;
+  private final BackOff exponentialBackOff;
+
+  private int currentRetries = 0;
+
+  public static UnsuccessfulResponseHandler newUnsuccessfulResponseHandler() {
+    return new UnsuccessfulResponseHandler(Sleeper.DEFAULT, new ExponentialBackOff());
+  }
+
+  @VisibleForTesting
+  UnsuccessfulResponseHandler(Sleeper sleeper, ExponentialBackOff exponentialBackOff) {
+    this.sleeper = sleeper;
+    this.exponentialBackOff = exponentialBackOff;
+  }
+
+  @Override
+  public boolean handleResponse(HttpRequest request, HttpResponse response, boolean supportsRetry)
+      throws IOException {
+    if (!supportsRetry) {
+      return false;
+    }
+    if (currentRetries >= MAX_RETRIES) {
+      return false;
+    }
+    currentRetries++;
+
+    var statusCode = response.getStatusCode();
+    // we only retry 5XX and 429
+    if (!(statusCode / 100 == 5 || statusCode == 429)) {
+      return false;
+    }
+    try {
+      var headers = response.getHeaders();
+      if (headers != null) {
+        var retryAfter = headers.getRetryAfter();
+        // we only use Retry-After on 503 and 429
+        if (retryAfter != null && (statusCode == 429 || statusCode == 503)) {
+          try {
+            return handleRetryAfter(retryAfter);
+          } catch (NumberFormatException ignored) {
+            // ignored this is a parse error from Retry-After, just go to exponential backoff
+            // but maybe pop off a warning so we can inform the infrastructure owners
+            log.warning(
+                "Retry-After header in request to "
+                    + request.getUrl()
+                    + " was invalid ("
+                    + retryAfter
+                    + ")");
+          }
+        }
+      }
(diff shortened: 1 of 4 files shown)


vlsi:
Could you please consider `Retry-After` header? (it might be in seconds or as HTTP date)

loosebazooka:
hrmm... gonna need to re-implement the whole handler.

loosebazooka:
I think we're mostly safe from that? The code to renew is before every rekor request.

vlsi:
Well, the code does try keeping some slack (`minSigningCertificateLifetime`, default `5 min`).
However, it might be better to reduce `minSigningCertificateLifetime` and handle failures from Rekor better.
AFAIK the current certificate lifetime is ~10 min, so if we start discarding certificates if they have less than 5 min left, then we use only 50% of the certificate :)

However, it is a different issue from handling 429

loosebazooka:
okay, well turns out this was a bit annoying (and the testing), but here we go.

sigstore-java/src/main/java/dev/sigstore/http/UnsuccessfulResponseHandler.java
```diff
@@ -0,0 +1,118 @@
...
+  @Override
+  public boolean handleResponse(HttpRequest request, HttpResponse response, boolean supportsRetry)
+      throws IOException {
+    if (!supportsRetry) {
+      return false;
+    }
+    try {
+      // exponential backoff for 5XX
+      if (response.getStatusCode() / 100 == 5) {
+        return BackOffUtils.next(sleeper, exponentialBackOff);
```
vlsi, line 67: `Retry-After` can easily appear in 503, so it would probably be nice to handle it as well

sigstore-java/src/main/java/dev/sigstore/http/UnsuccessfulResponseHandler.java
```diff
@@ -0,0 +1,118 @@
...
+      return date.toInstant().toEpochMilli() - now.toEpochMilli();
+    }
+
+    // maybe it is seconds
+    try {
+      return Long.parseLong(retryAfter) * 1000;
+    } catch (NumberFormatException nfe) {
+      // ignored
+    }
+    throw new IOException("Could not parse Retry-After");
```
vlsi, line 116: Should we just ignore unparseable `retry-after` instead of throwing an exception?

sigstore-java/src/main/java/dev/sigstore/http/UnsuccessfulResponseHandler.java
```diff
@@ -0,0 +1,118 @@
...
+ */
+public class UnsuccessfulResponseHandler implements HttpUnsuccessfulResponseHandler {
+  private final Sleeper sleeper;
+
+  // for 5XX
+  private final BackOff exponentialBackOff;
+
+  // for 429
+  private static final int MAX_429_RETRIES = 10;
+  private int current429Retries = 0;
```
vlsi, line 46: Is there a reason the "max retries" limit is only for 429?

loosebazooka: It's because the exponential back off has built in retires, but now that I think of it, I can just make that bigger and use a single count everywhere

sigstore-java/src/main/java/dev/sigstore/http/UnsuccessfulResponseHandler.java
loosebazooka: Oh right. Kk lemme work that in

sigstore-java/src/main/java/dev/sigstore/http/UnsuccessfulResponseHandler.java
loosebazooka: I'm not sure exactly what to do there. I don't actually expect to get there unless rekor is misbehaving. Ignoring it would mean we wouldn't detect it? Maybe a warn log and delegate to exponential back off?

vlsi: If the number is unparseable, then just delegate to regular exponential back off as if `retry-after` was missing.

loosebazooka: Yup

sigstore-java/src/main/java/dev/sigstore/http/UnsuccessfulResponseHandler.java
```diff
@@ -0,0 +1,123 @@
...
+    }
+    try {
+      if (currentRetries >= MAX_RETRIES) {
+        return false;
+      }
+      currentRetries++;
+
+      var statusCode = response.getStatusCode();
+      // we only retry 5XX and 429
+      if (statusCode / 100 == 5 || statusCode == 429) {
```
vlsi, line 73: I would prefer early return in case status code does not match.

It would reduce the nesting level

vlsi (approved):

vlsi (approved):

Issue #332, opened 2023-02-06
Release sigstore-java 0.3.0


loosebazooka:
- [x] tag and build 0.3.0 
    - https://github.com/sigstore/sigstore-java/actions/runs/4107271348
    - https://github.com/sigstore/sigstore-java/releases/tag/v0.3.0
- [x] release to maven central

FYI @patflynn @vlsi, tagging off of main@ 4ffe2674e1e9e268c00a4f4afa5264fdd399d453

loosebazooka:
This is a less than ideal publishing system, but ok to release (@patflynn, @vlsi)?

vlsi:
The more releases the better

PR #333, opened 2023-02-06
Fix reference to GitHub repo in release notes

szpak:
As it was broken in [0.3.0](https://github.com/sigstore/sigstore-java/releases/tag/v0.3.0) - `https://github.com/$GITHUB_REPOSITORY/CHANGELOG.md`. Related to #332.

Btw, it should work, but the release workflow is not available from a fork.


loosebazooka:
good call, we don't really even have a changelog yet though.

szpak:
> good call, we don't really even have a changelog yet though.

I've just found the same trying to read what have changed :D

loosebazooka (approved):

PR #334, opened 2023-02-06
Create CHANGELOG.md

loosebazooka:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,7 @@
+# Changelog
+
+All notable changes to `sigstore-java` will be documented in this file.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
+
+All versions prior to 1.0.0 are untracked


vlsi (approved):
I'm ok with this, however, I think release-drafter GitHub action + GitHub releases would be easier to maintain.

szpak:
> All versions prior to 1.0.0 are untracked

In fact, I believe that **important** changes should be documented, even in the releases prior to 1.0.0. Otherwise, the early adopters could be surprised by some big and non-backward compatible changes. Of course that can happen before 1.0.0, but in a good style is to notify about it.

szpak:
> I'm ok with this, however, I think release-drafter GitHub action + GitHub releases would be easier to maintain.

@vlsi Do you think about some "GitHub action" to generate the release notes based on the PRs with a specific labels (e.g. "major")?

loosebazooka:
github has some shortcuts to fill in the release notes into a release (since the last release or something).

vlsi:
See https://github.com/release-drafter/release-drafter

What it does it updates "draft release notes on GitHub" every time a PR is merged or updated.

It enables to customize release sections based on the PR labels, files modified, etc.

In my experience, 'keep a changelog' is pure pain

loosebazooka:
I'm okay changing this in the future, I'm going to merge it in for now.

PR #336, opened 2023-02-07
Update tag-and-build-release.yaml

loosebazooka:

just a minor update to #333


szpak (approved):
Definitely it's a better path :). As the file was not available before I didn't noticed that.. (although I could test with some other existing file anyway...)

vlsi (approved):

PR #344, opened 2023-02-14
Add all env to forked sigstore-java process

### LINKED ISSUE #337: process isolation in worker mechanism disables browser window from appearing
I've isolated the issue with local testing, but I think it actually holds true for anyone potentially using the gradle plugin that a browser wont pop up:

https://github.com/sigstore/sigstore-java/blob/main/sigstore-gradle/sigstore-gradle-sign-base-plugin/src/main/kotlin/dev/sigstore/sign/tasks/SigstoreSignFilesTask.kt#L196

It appears to have something to do with process isolation, if this is switched back to classpath isolation, things work as intended and the browser pops up (obviously with other oom concerns).

FYI @vlsi

loosebazooka:
fixes #337


vlsi (approved):
It would be nice if you could mention which var was missing in your case

loosebazooka:
it's what is needed to satisfy java.awt.Desktop [#isDesktopSupported​()](https://docs.oracle.com/javase/9/docs/api/java/awt/Desktop.html#isDesktopSupported--) and [#isSupported(Action.BROWSE)](https://docs.oracle.com/javase/9/docs/api/java/awt/Desktop.html#isSupported-java.awt.Desktop.Action-)

PR #345, opened 2023-02-17
refactor to use consistent snapshots

patflynn:
WIP towards #123 .  This changes the TUF updater to use consistent snapshots which is a change that was made to the TUF repo a few months ago.


sigstore-java/src/main/java/dev/sigstore/tuf/HttpMetaFetcher.java
```diff
@@ -58,15 +60,22 @@ public Optional<MetaFetchResult<Root>> getRootAtVersion(int version)
...
   }
 
   @Override
   public <T extends SignedTufMeta> Optional<MetaFetchResult<T>> getMeta(
-      Role.Name role, Class<T> t, Integer maxSize)
+      Role.Name role, int version, Class<T> t, Integer maxSize)
       throws IOException, FileExceedsMaxLengthException {
-    String fileName = role.name().toLowerCase(Locale.ROOT) + ".json";
-    return getMeta(fileName, t, maxSize);
+    Preconditions.checkArgument(version > 0, version + " should be a positive value.");
```
vlsi, line 70: Prefer template-based API to avoid string concatenation when the check passes: https://guava.dev/releases/19.0/api/docs/com/google/common/base/Preconditions.html#checkArgument(boolean,%20java.lang.String,%20java.lang.Object...)

```suggestion
    Preconditions.checkArgument(version > 0, "version should be positive, got: %d", version);
```

vlsi (dismissed):
LGTM modulo a small thing re Preconditions

vlsi (approved):

sigstore-java/src/main/java/dev/sigstore/tuf/HttpMetaFetcher.java
patflynn: good call. fixed.

PR #351, opened 2023-02-19
test: distinguish between locally installed 0.3 and remote 0.3 sigstore-java better

vlsi:
#### Summary

Previously the test was comparing "requested version with project version" to tell if it needs to fetch sigstore-java from Central.
Now there is `LocallyBuiltVersion` class so the test can distinguish better, especially when project version in `gradle.properties` was not updated after release.

#### Release Note

NONE

#### Documentation

NONE


loosebazooka:
Thanks, I should probably just do deprecation and releases in a more structured manner, even if we are not GA.

loosebazooka (approved):

PR #366, opened 2023-03-08
Add parsing of bundle to KeylessSigningResult

loosebazooka:
Should probably rename KeylessSigningResult to something else (KeylessSignature or something). But that's for another day.

### DIFF
--- a/sigstore-java/src/main/java/dev/sigstore/bundle/BundleFactory.java
+++ b/sigstore-java/src/main/java/dev/sigstore/bundle/BundleFactory.java
@@ -18,6 +18,7 @@ package dev.sigstore.bundle;
 import com.google.protobuf.InvalidProtocolBufferException;
 import dev.sigstore.KeylessSigningResult;
 import dev.sigstore.proto.bundle.v1.Bundle;
+import java.io.Reader;
 import java.util.List;
 
 /**
@@ -50,5 +51,18 @@ public class BundleFactory {
       throw new IllegalArgumentException(
           "Can't serialize signing result to Sigstore Bundle JSON", e);
     }
+  }
+
+  /**
+   * Read a bundle json and convert it back into a keyless signing result for use within this
+   * library
+   *
+   * @param jsonReader a reader to a valid bundle json file
+   * @return the converted signing result object
+   * @throws BundleParseException if all or parts of the bundle were not convertible to library
+   *     types
+   */
+  public static KeylessSigningResult readBundle(Reader jsonReader) throws BundleParseException {
+    return BundleFactoryInternal.readBundle(jsonReader);
   }
 }
--- a/sigstore-java/src/main/java/dev/sigstore/bundle/BundleFactoryInternal.java
+++ b/sigstore-java/src/main/java/dev/sigstore/bundle/BundleFactoryInternal.java
@@ -17,15 +17,39 @@ package dev.sigstore.bundle;
 
 import com.google.protobuf.ByteString;
 import com.google.protobuf.util.JsonFormat;
+import dev.sigstore.ImmutableKeylessSigningResult;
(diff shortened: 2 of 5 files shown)


sigstore-java/src/main/java/dev/sigstore/bundle/BundleFactory.java
```diff
@@ -51,4 +54,19 @@ public static String createBundle(KeylessSigningResult signingResult) {
...
+   * Read a bundle json and convert it back into a keyless signing result for use within this
+   * library
+   *
+   * @param jsonReader a reader to a valid bundle json file
+   * @return the converted signing result object
+   */
+  public static KeylessSigningResult readBundle(Reader jsonReader) {
+    try {
+      return BundleFactoryInternal.readBundle(jsonReader);
+    } catch (IOException | CertificateException e) {
```
patflynn, line 68: should the method just throw IOException?

loosebazooka: good point, added BundleParseException

sigstore-java/src/main/java/dev/sigstore/bundle/BundleFactory.java
```diff
@@ -51,4 +52,17 @@ public static String createBundle(KeylessSigningResult signingResult) {
...
+  /**
+   * Read a bundle json and convert it back into a keyless signing result for use within this
+   * library
+   *
+   * @param jsonReader a reader to a valid bundle json file
+   * @return the converted signing result object
+   * @throws BundleParseException if all or parts of the bundle were not convertible to library
+   *     types
+   */
+  public static KeylessSigningResult readBundle(Reader jsonReader) throws BundleParseException {
```
patflynn, line 65: Why not just take String like the createBundle method produces?

patflynn (dismissed):
as an aside I'm not familiar with the public passthrough to a *internal class, Kinda curious what purpose what the pattern serves.

sigstore-java/src/main/java/dev/sigstore/bundle/BundleFactoryInternal.java
```diff
@@ -120,4 +144,79 @@ private static void addInclusionProof(
...
+            .logID(Hex.toHexString(bundleEntry.getLogId().getKeyId().toByteArray()))
+            .logIndex(bundleEntry.getLogIndex())
+            .body(
+                Base64.getEncoder()
+                    .encodeToString(bundleEntry.getCanonicalizedBody().toByteArray()))
+            .verification(verification)
+            .build();
+
+    if (bundle.getMessageSignature().getMessageDigest().getAlgorithm() != HashAlgorithm.SHA2_256) {
+      throw new BundleParseException("Cannot read message signatures that are not sha256");
```
vlsi, line 197: We should probably add actual algorithm name into the error message

loosebazooka:
When the internal class was made, I believe @vlsi's intention was to isolate the requirement of pulling in the protobuf (so it could be split out into another library in the future). So I just kinda followed it. If we don't see a need for it, we can move it in.

sigstore-java/src/main/java/dev/sigstore/bundle/BundleFactory.java
loosebazooka: Because it could be coming from a file. Using a string is just as easy as doing `new StringReader("my-string")'` which is what happens internally in the library (if you use the `merge(String...)` method)

sigstore-java/src/main/java/dev/sigstore/bundle/BundleFactoryInternal.java
loosebazooka: turns out there isn't much else (just sha2_256, unspecified and unrecognized), but added it in the message for the future.

PR #367, opened 2023-03-09
fuzzing: Add 3 fuzzers and OSS-Fuzz build script

AdamKorcz:
#### Summary

Adds three fuzzers and the OSS-Fuzz build script.

All work made by @arthurscchan

### DIFF
--- a/fuzzing/oss_fuzz_build.sh
+++ b/fuzzing/oss_fuzz_build.sh
@@ -1,0 +1,50 @@
+#!/bin/bash -eu
+
+# TODO AdamKorcz: Make the build script a gradle module
+
+./gradlew clean build -x test --refresh-dependencies
+
+CURRENT_VERSION=$(./gradlew properties --no-daemon --console=plain -q | grep "^version:" | awk '{printf $2}')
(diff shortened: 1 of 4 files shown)


fuzzing/oss_fuzz_build.sh
```diff
@@ -0,0 +1,48 @@
+#!/bin/bash -eu
```
loosebazooka, line 1: It actually looks like to me that this could just be it's own gradle module with a dependency on sigstore-java? Delegate to gradle to actually do the right thing for the build?

loosebazooka: I don't know how the jazzer_api is injected in (presumably in the build container?)

loosebazooka: you could use the application plugin to create runnable targets?

AdamKorcz: I think it could be a gradle module as long as we get the flags from the OSS-Fuzz environment. I am happy to make it a TODO, but would prefer to get the fuzzers running the current way first. Is the current build script a blocker to you?

loosebazooka: kk sounds good.

loosebazooka (approved):

PR #372, opened 2023-03-14
refactor fuzzing into a gradle sub-project

loosebazooka:
There's probably a bit more to do here, we can try using the jazzer junit project to orchestate the fuzzing, but this is just a step in that direction

FYI @AdamKorcz not sure how to go about testing this though, maybe we should add a readme to the fuzzing subproject

### DIFF
--- a/fuzzing/build.gradle.kts
+++ b/fuzzing/build.gradle.kts
@@ -1,0 +1,17 @@
+plugins {
+    id("build-logic.java")
+}
+
+repositories {
+    mavenCentral()
+}
+
+dependencies {
+    implementation(project(":sigstore-java"))
+    implementation("com.code-intelligence:jazzer-api:0.16.0")
+}
+
+tasks.register<Copy>("copyToFuzzOut") {
+    dependsOn(tasks.build)
+    into(project.property("fuzzOut")).from(sourceSets.main.get().runtimeClasspath)
+}
--- a/fuzzing/oss_fuzz_build.sh
+++ b/fuzzing/oss_fuzz_build.sh
@@ -1,37 +1,27 @@
 #!/bin/bash -eu
 
-# TODO AdamKorcz: Make the build script a gradle module
+# TODO: this should be a gradle plugin
 
-./gradlew clean build -x test --refresh-dependencies
-
-CURRENT_VERSION=$(./gradlew properties --no-daemon --console=plain -q | grep "^version:" | awk '{printf $2}')
-cp sigstore-java/build/libs/sigstore-java-$CURRENT_VERSION.jar $OUT/sigstore-java.jar
-
-# Search for dependency jars from gradle local repository,
-# move them to the $OUT directory and add them to the ALL_JARS
-# variable to allow jvm to use them as class path when compiling
-# and executing the fuzzers which depends on them
-ALL_JARS=sigstore-java.jar
-for jarfile in $(find ~/.gradle/caches/modules-2/files-2.1/ \( -iname "*.jar" ! -iname "junit*" ! -iname "spotless*" ! -iname "ktlint*" ! -iname "gradle*" ! -iname "kotlin*" \))
+# build the fuzzing classes and extract dependencies into $OUT
+./gradlew :fuzzing:copyToFuzzOut -x test -PfuzzOut="$OUT"
+ALL_JARS=""
+for jarfile in $(find $OUT -name *.jar)
 do
-  cp $jarfile $OUT/
   ALL_JARS="$ALL_JARS $(basename $jarfile)"
 done
-
-# The classpath at build-time includes the project jars in $OUT as well as the
-# Jazzer API.
-BUILD_CLASSPATH=$(echo $ALL_JARS | xargs printf -- "$OUT/%s:"):$JAZZER_API_PATH
 
 # All .jar and .class files lie in the same directory as the fuzzer at runtime.
 RUNTIME_CLASSPATH=$(echo $ALL_JARS | xargs printf -- "\$this_dir/%s:"):\$this_dir
 
-for fuzzer in $(find $SRC/sigstore-java/fuzzing -name '*Fuzzer.java'); do
-  fuzzer_basename=$(basename -s .java $fuzzer)
+# Create all fuzzing targets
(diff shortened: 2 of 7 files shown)


loosebazooka:
@AdamKorcz I think this is ready to merge, so we can use fuzzing as a gradle project

loosebazooka:
fyi @arthurscchan lemme know what you think?

loosebazooka:
I think a very long term goal for java should be a oss-fuzz plugin that does all of this for the user instead of having them write bash to do it.

fuzzing/oss_fuzz_build.sh
```diff
@@ -1,37 +1,30 @@
...
-cp sigstore-java/build/libs/sigstore-java-$CURRENT_VERSION.jar $OUT/sigstore-java.jar
-
-# Search for dependency jars from gradle local repository,
-# move them to the $OUT directory and add them to the ALL_JARS
-# variable to allow jvm to use them as class path when compiling
-# and executing the fuzzers which depends on them
-ALL_JARS=sigstore-java.jar
-for jarfile in $(find ~/.gradle/caches/modules-2/files-2.1/ \( -iname "*.jar" ! -iname "junit*" ! -iname "spotless*" ! -iname "ktlint*" ! -iname "gradle*" ! -iname "kotlin*" \))
+# build the fuzzing classes and extract dependencies into the build/fuzzRoot
+./gradlew :fuzzing:build :fuzzing:copyToFuzzOut -x test --refresh-dependencies -PfuzzOut="$OUT"
```
szpak, line 4: I realize, the behavior was the same before this change, but maybe the script could check for the existence of `$OUT` and report clear error that it has to be set?

2. I wonder, if `-x test` is needed? It failed due to the missing `src/test/*`?

3. Why `--refresh-dependencies`? Are there any SNAPSHOTs used?

szpak (commented):
+1 for having some README - how to use it.

fuzzing/oss_fuzz_build.sh
loosebazooka: 1. $Out appears to come from the fuzzing infra, so out looks like it *must* be there
2. I think we don't care about running tests. We're just building to get the fuzzers going.
3. I didn't see that, I would assume that it shouldn't matter since all builds are fresh?

arthurscchan: 1. $OUT is defined in the fuzzing infra while running the fuzzer in a docker image. So it must be there.
2. Yes, we don't care about running tests because in theory fuzzer are also a kind of test on its own. The exclude of test may speed up the process.
3. Yes, it may not necessary, the original reason for adding that is to make sure the fuzzer could retrieve an up to date dependency jar for execution.

DavidKorczynski:
I think having it as a sub-project makes sense -- At the moment, we're seeing some issues where some dependencies are not being correctly uploaded with the fuzzing jobs e.g. https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=57189 which causes the fuzzer to not run. I think this is a first thing we need to resolve to confirm things are running proper -- @arthurscchan pointed out this PR may resolve it although we're not entirely sure yet -- I'll see if I can do some debugging today on this.

Another thing we should do in this context is add a CIFuzz job as it'll help test fuzzing changes. @AdamKorcz  could we set this up?

AdamKorcz:
> I think having it as a sub-project makes sense -- At the moment, we're seeing some issues where some dependencies are not being correctly uploaded with the fuzzing jobs e.g. https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=57189 which causes the fuzzer to not run. I think this is a first thing we need to resolve to confirm things are running proper -- @arthurscchan pointed out this PR may resolve it although we're not entirely sure yet -- I'll see if I can do some debugging today on this.
> 
> Another thing we should do in this context is add a CIFuzz job as it'll help test fuzzing changes. @AdamKorcz could we set this up?

Yes, CIFuzz has been set up here: https://github.com/sigstore/sigstore-java/pull/381

DavidKorczynski:
> At the moment, we're seeing some issues where some dependencies are not being correctly uploaded with the fuzzing jobs e.g. https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=57189 which causes the fuzzer to not run. I think this is a first thing we need to resolve to confirm things are running proper -- @arthurscchan pointed out this PR may resolve it although we're not entirely sure yet -- I'll see if I can do some debugging today on this.

For reference, this should now be fixed by: https://github.com/sigstore/sigstore-java/pull/383

fuzzing/oss_fuzz_build.sh
szpak: > $OUT is defined in the fuzzing infra while running the fuzzer in a docker image. So it must be there.

Ok. I was rather thinking about the people willing to run it locally (e.g. to verify the changes they made in the scripts or configuration). So probably, some extra check could be in place anyway. However, due to the missing "README", I even not sure, if it is "easy" to run it locally (even to play with for some short time) or rather the CI action is more feasible in general :-/.

loosebazooka:
okay y'all I think this is ready again. I do think the right thing to do eventually here is generate all of this from gradle directly at some point -- https://github.com/sigstore/sigstore-java/pull/372#issuecomment-1474354552

AdamKorcz:
I have tested this locally and it runs well. I would be happy to see this merged and see how it runs in prod on OSS-Fuzz.

loosebazooka:
@AdamKorcz that's good news, lemme get it merged and maybe we can work a gradle plugin to do all the heavy lifting

fuzzing/build.gradle.kts
```diff
@@ -0,0 +1,17 @@
...
+}
+
+dependencies {
+    implementation(project(":sigstore-java"))
+    implementation("com.code-intelligence:jazzer-api:0.16.0")
+}
+
+tasks.register<Copy>("copyToFuzzOut") {
+    dependsOn(tasks.build)
+    into(project.property("fuzzOut")).from(sourceSets.main.get().runtimeClasspath)
```
vlsi, line 16: It would probably be better to use a fixed output here.
Currently this code breaks `./gradlew tasks` execution with https://github.com/sigstore/sigstore-java/issues/412

loosebazooka: this output needs to be to a specific location for fuzzing, but I added a default so this doesn't break stuff

PR #375, opened 2023-03-15
test: test sigstore-gradle plugins only with the locally built sigstore-java

vlsi:
#### Summary

It might be fine for now if sigstore-gradle plugin requires exactly the same sigstore-java version.

#### Release Note

NONE

#### Documentation

NONE


loosebazooka (approved):
might as well re-enable the tests then?
SigstoreSignTest.kt 
SigstorePublishSignTest.kt

loosebazooka (approved):

PR #377, opened 2023-03-16
Add verification on supplied artifact hash

loosebazooka:
It's possible someone supplied a bundle without verifying it's the same hash as the artifact.

### DIFF
--- a/sigstore-java/src/main/java/dev/sigstore/KeylessVerifier.java
+++ b/sigstore-java/src/main/java/dev/sigstore/KeylessVerifier.java
@@ -16,6 +16,8 @@
 package dev.sigstore;
 
 import com.google.api.client.util.Preconditions;
+import com.google.common.hash.Hashing;
+import com.google.common.io.Files;
 import dev.sigstore.KeylessVerificationRequest.VerificationOptions;
 import dev.sigstore.encryption.certificates.Certificates;
 import dev.sigstore.encryption.signers.Verifiers;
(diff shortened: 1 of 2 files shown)


sigstore-java/src/main/java/dev/sigstore/KeylessVerifier.java
```diff
@@ -132,10 +137,38 @@ public void verifyOnline(byte[] artifactDigest, byte[] certChain, byte[] signatu
...
+  public void verify(byte[] artifactDigest, KeylessVerificationRequest request)
+      throws KeylessVerificationException {
     var signingCert = SigningCertificate.from(request.getKeylessSignature().getCertPath());
     var leafCert = signingCert.getLeafCertificate();
 
+    // this ensures the provided artifact digest matches what may have come from a bundle (in
+    // keyless signature)
+    if (!Arrays.equals(artifactDigest, request.getKeylessSignature().getDigest())) {
+      throw new KeylessVerificationException(
+          "Provided artifact digest does not match digest used for verification");
```
vlsi, line 169: We should probably list both hashes for the user to analyze

vlsi (changes requested):

sigstore-java/src/main/java/dev/sigstore/KeylessVerifier.java
loosebazooka: done

PR #378, opened 2023-03-16
fix build, merges went weird

loosebazooka:


loosebazooka:
unfortunately conformance is running on head as that is how `pull-request-target` works, but here are the result of conformance off of the branch: https://github.com/sigstore/sigstore-java/actions/runs/4440369664

vlsi (approved):

loosebazooka:
merged after bypassing conformance checks, but those should be fixed now anyway.

PR #387, opened 2023-03-21
Encoder: Add missing exception handling

arthurscchan:
#### Summary

OSS-fuzz integration finds a missing exception handling. The Base64 decoder throws an DecoderException for malformed PEM string but the current code does not catch and handle it. This PR adds a try catch block to wrap and throw the exception. 
Fixes: https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=57244

### DIFF
--- a/sigstore-java/src/main/java/dev/sigstore/encryption/Keys.java
+++ b/sigstore-java/src/main/java/dev/sigstore/encryption/Keys.java
@@ -71,7 +71,7 @@ public class Keys {
             "sigstore public keys must be only a single PEM encoded public key");
       }
     } catch (DecoderException e) {
-      throw new InvalidKeySpecException("Invalid key");
+      throw new InvalidKeySpecException("Invalid key, could not parse PEM section");
     }
     // special handling for PKCS1 (rsa) public key
     if (section == null) {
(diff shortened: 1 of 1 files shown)


sigstore-java/src/main/java/dev/sigstore/encryption/Keys.java
```diff
@@ -62,10 +63,15 @@ public static PublicKey parsePublicKey(byte[] keyBytes)
...
-          "sigstore public keys must be only a single PEM encoded public key");
+    PemObject section = null;
+    try {
+      section = pemReader.readPemObject();
+      if (pemReader.readPemObject() != null) {
+        throw new InvalidKeySpecException(
+            "sigstore public keys must be only a single PEM encoded public key");
+      }
+    } catch (DecoderException e) {
+      throw new InvalidKeySpecException("Invalid key");
```
loosebazooka, line 74:
```suggestion
      throw new InvalidKeySpecException("Invalid key, could not parse PEM section");
```

loosebazooka (dismissed):
Can we add a little more info to the error?

sigstore-java/src/main/java/dev/sigstore/encryption/Keys.java
arthurscchan: Thanks for the comments. Changed accordingly.

loosebazooka (approved):

PR #394, opened 2023-03-22
Exception fixing: Add handling for possible empty content for PemObject

arthurscchan:
#### Summary

PemObject may contain empty content and result in NullPointerException when its content is being used as an argument for creating a public key. This PR adds a check to ensure PemObject content is not null before use.

### DIFF
--- a/sigstore-java/src/main/java/dev/sigstore/encryption/Keys.java
+++ b/sigstore-java/src/main/java/dev/sigstore/encryption/Keys.java
@@ -75,7 +75,7 @@ public class Keys {
     }
     // special handling for PKCS1 (rsa) public key
     if ((section == null) || (section.getContent() == null)) {
-      throw new InvalidKeySpecException("Invalid key");
+      throw new InvalidKeySpecException("Invalid key, empty PEM section");
     }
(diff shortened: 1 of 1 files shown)


sigstore-java/src/main/java/dev/sigstore/encryption/Keys.java
```diff
@@ -74,7 +74,7 @@ public static PublicKey parsePublicKey(byte[] keyBytes)
       throw new InvalidKeySpecException("Invalid key, could not parse PEM section");
     }
     // special handling for PKCS1 (rsa) public key
-    if (section == null) {
+    if ((section == null) || (section.getContent() == null)) {
       throw new InvalidKeySpecException("Invalid key");
```
loosebazooka, line 78: while we're in here, can we just add in "Invalid key, empty section" or something to the message?

loosebazooka (approved):

sigstore-java/src/main/java/dev/sigstore/encryption/Keys.java
arthurscchan: Changed accordingly.

loosebazooka (approved):

PR #404, opened 2023-03-29
Fix: Fix possible Null Pointer Exception

arthurscchan:
#### Summary

In [PublicKeyFactory Line 108](https://github.com/bcgit/bc-java/blob/master/core/src/main/java/org/bouncycastle/crypto/util/PublicKeyFactory.java#L108), SubjectPublicKeyInfo.getInstance method could return null and make [PublicKeyFactory Line 148](https://github.com/bcgit/bc-java/blob/master/core/src/main/java/org/bouncycastle/crypto/util/PublicKeyFactory.java#L148) throws a NullPointerException. 

From [ASN1Primitive](https://github.com/bcgit/bc-java/blob/master/core/src/main/java/org/bouncycastle/asn1/ASN1Primitive.java#L44-L51) and [ASN1InputStream](https://github.com/bcgit/bc-java/blob/master/core/src/main/java/org/bouncycastle/asn1/ASN1InputStream.java#L189-L198), we know that if the tag (first  byte of the provided byte array) is a number less than 0, it will return null value and cause the null pointer exception. Thus the fix is to check if the first byte is less than or equals to 0 (it will throw IOException if the number is equals to 0) to ensure the SubjectPublicKeyInfo object is not null when creating the key.

### DIFF
--- a/sigstore-java/src/main/java/dev/sigstore/encryption/Keys.java
+++ b/sigstore-java/src/main/java/dev/sigstore/encryption/Keys.java
@@ -94,6 +94,11 @@ public class Keys {
 
     // Ensure PEM content can be parsed correctly
     try {
+      // First byte of the content byte array is the tag and
+      // cannot be less than or equals to 0
+      if (byte[0] <= 0) {
+        throw new InvalidKeySpecException("Invalid key, could not parse PEM content");
+      } 
       keyParameters = PublicKeyFactory.createKey(content);
     } catch (IllegalStateException e) {
       throw new InvalidKeySpecException("Invalid key, could not parse PEM content");


vlsi:
I would disagree with the analysis.

`byte` is signed in Java, so there's nothing wrong with having "negative" bytes.

I see you mention https://github.com/bcgit/bc-java/blob/master/core/src/main/java/org/bouncycastle/asn1/ASN1InputStream.java#L189-L198, however, it refers to `int` result of `.read()`, see https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/io/FilterInputStream.html#read()

arthurscchan:
Sorry for the wrong interpretation.

PR #405, opened 2023-03-29
Fixes: Add digest length checking

arthurscchan:
According to the [ECDSASignature class](https://github.com/AdoptOpenJDK/openjdk-jdk11/blob/master/src/jdk.crypto.ec/share/classes/sun/security/ec/ECDSASignature.java#L127), the NoneWithECDSA mode, the maximum length for the digest is 64, using a digest longer than 64 will result in ArrayIndexOutOfBoundException when the full byte array is copied to a byte array with only 64 bytes.

Fixes: https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=57360

### DIFF
--- a/sigstore-java/src/main/java/dev/sigstore/encryption/signers/EcdsaSigner.java
+++ b/sigstore-java/src/main/java/dev/sigstore/encryption/signers/EcdsaSigner.java
@@ -43,6 +43,9 @@ public class EcdsaSigner implements Signer {
   @Override
   public byte[] signDigest(byte[] artifactDigest)
       throws NoSuchAlgorithmException, InvalidKeyException, SignatureException {
+    if (artifactDigest.length > 64) {
+      throw new SignatureExeption("Artifact digest cannot be longer than 64 bytes for this mode");
+    }
     Signature signature = Signature.getInstance("NONEwithECDSA");
     signature.initSign(keyPair.getPrivate());
     signature.update(artifactDigest);
--- a/sigstore-java/src/main/java/dev/sigstore/encryption/signers/EcdsaVerifier.java
+++ b/sigstore-java/src/main/java/dev/sigstore/encryption/signers/EcdsaVerifier.java
@@ -43,6 +43,9 @@ public class EcdsaVerifier implements Verifier {
   @Override
   public boolean verifyDigest(byte[] digest, byte[] signature)
       throws NoSuchAlgorithmException, InvalidKeyException, SignatureException {
+    if (digest.length > 64) {
+      throw new SignatureExeption("Artifact digest cannot be longer than 64 bytes for this mode");
+    }
     var verifier = Signature.getInstance("NONEwithECDSA");
     verifier.initVerify(publicKey);
     verifier.update(digest);


sigstore-java/src/main/java/dev/sigstore/encryption/signers/EcdsaSigner.java
```diff
@@ -43,6 +43,9 @@ public byte[] sign(byte[] artifact)
   @Override
   public byte[] signDigest(byte[] artifactDigest)
       throws NoSuchAlgorithmException, InvalidKeyException, SignatureException {
+    if (artifactDigest.length > 64) {
+      throw new SignatureExeption("Artifact digest cannot be longer than 64 bytes for this mode");
+    }
```
vlsi, line 48: How is this better than the exception from `signature.update(artifactDigest);` ?

DavidKorczynski: It's better in that it's a controlled exception with a meaningful description attached to it.

The exception thrown is:
```
== Java Exception: java.lang.ArrayIndexOutOfBoundsException: arraycopy: last destination index 65 out of bounds for byte[64]
 at java.base/java.lang.System.arraycopy(Native Method)
 at jdk.crypto.ec/sun.security.ec.ECDSASignature$RawECDSA.engineUpdate(ECDSASignature.java:158)
 at java.base/java.security.Signature$Delegate.engineUpdate(Signature.java:1414)
 at java.base/java.security.Signature.update(Signature.java:903)
 at java.base/java.security.Signature.update(Signature.java:872)
 at dev.sigstore.encryption.signers.EcdsaSigner.signDigest(EcdsaSigner.java:48)
 at fuzzing.SignerVerifierFuzzer.fuzzerTestOneInput(SignerVerifierFuzzer.java:37)
```

vlsi: I wonder if the check should be implemented in the downstream library instead. Of course, it is easier to sweep it under the rug here, however, it is hard to understand why we have an artificial limit of 64 here.

So I would suggest filing an issue/PR for the impl library rather than sigstore-java

DavidKorczynski: That makes sense to me. Let's make a PR there and see how it's retrieved.

loosebazooka: Just checking in? Was there any resolution on this? Is there a PR somewhere to follow up on?

DavidKorczynski: Ah, I missed this! Apologies -- @arthurscchan could you help with this one?

DavidKorczynski: I just spoke to @arthurscchan and @AdamKorcz who said it may not be the correct choice to change upstream. @AdamKorcz could you elaborate?

AdamKorcz: The best "upstream" place to fix this issue is at the provider level. However, this is not a suitable place to fix this issue, since there are many providers that sigstore-java users can consume and/or write their own providers. As such, sigstore-java is the best place to do the digest length check.

loosebazooka (dismissed):

loosebazooka:
ugh sorry, not all these statuses are manually triggerable. Can you rebase it so we can get all the statuses green?

loosebazooka:
looks like this needs some manual intervention

PR #406, opened 2023-03-29
Fix: Fix possible Null Pointer Exception

arthurscchan:
#### Summary

In [PublicKeyFactory Line 108](https://github.com/bcgit/bc-java/blob/master/core/src/main/java/org/bouncycastle/crypto/util/PublicKeyFactory.java#L108), SubjectPublicKeyInfo.getInstance method could return null and make [PublicKeyFactory Line 148](https://github.com/bcgit/bc-java/blob/master/core/src/main/java/org/bouncycastle/crypto/util/PublicKeyFactory.java#L148) throws a NullPointerException. 

From [InputStream javadoc](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/io/InputStream.html#read()), we see that if the end of the stream is reached, it will return -1. Which means that if an empty byte array is provided, the tag will be -1 and it will return null in [ASN1InputStream](https://github.com/bcgit/bc-java/blob/master/core/src/main/java/org/bouncycastle/asn1/ASN1InputStream.java#L189-L198). In this case, it will trigger the null pointer exception above. Thus this fix adds an additional checking to ensure the PEM content is not an empty array.

### DIFF
--- a/sigstore-java/src/main/java/dev/sigstore/encryption/Keys.java
+++ b/sigstore-java/src/main/java/dev/sigstore/encryption/Keys.java
@@ -74,7 +74,7 @@ public class Keys {
       throw new InvalidKeySpecException("Invalid key, could not parse PEM section");
     }
     // special handling for PKCS1 (rsa) public key
-    if ((section == null) || (section.getContent() == null)) {
+    if ((section == null) || (section.getContent() == null) || (section.getContent().length == 0)) {
       throw new InvalidKeySpecException("Invalid key, empty PEM section");
     }
     if (section.getType().equals("RSA PUBLIC KEY")) {


loosebazooka (dismissed):

sigstore-java/src/main/java/dev/sigstore/encryption/Keys.java
```diff
@@ -74,7 +74,7 @@ public static PublicKey parsePublicKey(byte[] keyBytes)
       throw new InvalidKeySpecException("Invalid key, could not parse PEM section");
     }
     // special handling for PKCS1 (rsa) public key
-    if ((section == null) || (section.getContent() == null)) {
+    if ((section == null) || (section.getContent() == null) || (section.getContent().length == 0)) {
```
vlsi, line 77: Why do we explicitly verify for 0-byte long array?
What if the array is just 1 byte?

vlsi (changes requested):
I think the verification should be performed by the downstream library, and adding 0 check here is wrong

loosebazooka:
@vlsi, I believe there is a sufficient description to describe the issue. The title is aligned with the problem.

vlsi:
@loosebazooka , we should not add arbitrary checks. Is there anything in ASN1Sequence.getInstance that says “empty byte array must not be passed there”?
If empty array triggers issues, it should be documented and tested at bc-java level, and we should not add those checks at random in sigatore-java.

Then, I do not like to merge changes without corresponding tests.
The PR add no tests, so it is possible that someone would remove the “useless” check tomorrow in a different PR.

loosebazooka:
I mean I agree here, this is a bouncy castle issue, but it still causes crashes in our code.
I believe oss-fuzz caches issues that cause crashes and you can repeat those against the proejct (is this right @arthurscchan ?) So if these were removed, oss-fuzz would catch it.

arthurscchan:
@loosebazooka Indeed, this could be reproducible.

loosebazooka:
@arthurscchan can you elaborate or point to the docs on how reproducing works for ossfuzz, it might also make sense to just bake these test cases into our testsuite anyway.

arthurscchan:
In general, the oss-fuzz tool provides function to reproduce testcase, see https://google.github.io/oss-fuzz/advanced-topics/reproducing/#reproduce-using-local-source-checkout.
The testcase that cause the problem could find in the bug page.

DavidKorczynski:
> In general, the oss-fuzz tool provides function to reproduce testcase, see https://google.github.io/oss-fuzz/advanced-topics/reproducing/#reproducing-bugs The testcase that cause the problem could find in the bug page.

In addition to this, we also added some documentation here: https://github.com/sigstore/sigstore-java/blob/main/DEVELOPMENT.md#reproduce-oss-fuzz-crash

To reproduce this crash:
```
git clone https://github.com/google/oss-fuzz
cd oss-fuzz
python3 infra/helper.py build_fuzzers sigstore-java
# From this page: https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=57247
# download `Reproducer Testcase: https://oss-fuzz.com/download?testcase_id=5658600826863616` to the current folder
python3 infra/helper.py reproduce sigstore-java KeysFuzzer ./clusterfuzz-testcase-minimized-KeysFuzzer-5658600826863616
```

loosebazooka:
I think it still makes sense to include a testcase in KeysTest.java to document this behavior. The relationship between ossfuzz and this project is not guaranteed to be transferable. Anyone forking this project or keeping a private copy should be allowed the benefit of a full testsuite.

DavidKorczynski:
> I think it still makes sense to include a testcase in KeysTest.java to document this behavior

@arthurscchan can you do this?

vlsi (changes requested):
Please structure tests in such a way that they fail before you apply the fix.
The current test passes even with the current `sigstore-java`, so it does not look like the test is helpful

sigstore-java/src/test/java/dev/sigstore/encryption/KeysTest.java
```diff
@@ -184,6 +185,18 @@ void parseTufPublicKey_rsa()
     }
   }
 
+  @Test
+  void parsePublicKeyBad()
+      throws IOException, InvalidKeySpecException, NoSuchAlgorithmException,
+          NoSuchProviderException {
+    try {
+      byte[] byteArray = "-----BEGIN A-----\nBBBBB-----END A".getBytes(Charset.defaultCharset());
```
loosebazooka, line 193: use `StandardCharsets.UTF_8`

sigstore-java/src/test/java/dev/sigstore/encryption/KeysTest.java
```diff
@@ -184,6 +185,18 @@ void parseTufPublicKey_rsa()
     }
   }
 
+  @Test
+  void parsePublicKeyBad()
```
loosebazooka, line 189: So yeah these kind of negative tests in junit5 should follow the pattern
```
  // add a comment to the link to oss-fuzz test case?
  @Test
  void parsePublicKey_failOnNullSection() {
    byte[] key = "-----BEGIN A-----\nBBBBB-----END A".getBytes(StandardCharsets.UTF_8);
    Assertions.assertThrows(InvalidKeySpecException.class, () -> Keys.parsePublicKey(key));
  }
```

loosebazooka: you can save the result from Assertions.assertThrows and compare the message in the Exception to what's expected too, but I don't think that's necessary here in this case.

arthurscchan: Thanks for the suggestion. I am thinking the opposite, we want the unit testing to assert that the NPE is not thrown with the fix, right? Thus Assertions.assertThrows may not be the one we want to use, instead the fail() should exist in the catch block if the exception caught is a NPE. In other words, I am assuming that the unit testing should fail for throwing NPE before the fix and pass after the fix.

arthurscchan: Thanks for the suggestion. Changed accordingly.

sigstore-java/src/test/java/dev/sigstore/encryption/KeysTest.java
arthurscchan: Changed accordingly. Thanks.

sigstore-java/src/test/java/dev/sigstore/encryption/KeysTest.java
```diff
@@ -184,6 +185,13 @@ void parseTufPublicKey_rsa()
...
```
loosebazooka, line 189: can you update naming here `parsePublicKey_failOnNullSection()`. If you could also point to the oss_fuzz failure in a comment, I think that makes this a bit easier to track to origin since the testcase data is a bit weird.

arthurscchan: Change accordingly, added a link to the oss-fuzz bug page in the comment.

loosebazooka (approved):

loosebazooka (approved):

vlsi:
The bug has nothing to do with sigstore-java, and the bug belongs to bc-java. If we commit a workaround, there should be a reference to bc-java issue, so we could drop the workaround once bc-java issue is resolved

loosebazooka:
Sure, we can link to https://github.com/bcgit/bc-java/issues/1370 for now.

vlsi (approved):
thanks

PR #407, opened 2023-03-29
Fuzzing: Add fuzzer for DigitallySigned class

arthurscchan:
#### Summary

Adding  fuzzer focusing on DigitallySigned classes.


DavidKorczynski:
> Fuzzing this is probably fine, but we shouldn't be the primary fuzzer for this conscrypt project. I guess we could always push our changes upstream?

In essence I would do both. We will look at integrating conscrypt -- but at the same time I would suggest analysing all code of sigstore-java.

loosebazooka (approved):

AdamKorcz (approved):
LGTM. IMO it is a good target given that sigstore-java has the code it its own repository.

PR #411, opened 2023-04-04
Use spec-compliant persisted target filenames

loosebazooka:
@asraa recently informed me that we shouldn't be persisting down to disk (or wherever) with the sha prefixed filename.

context: https://theupdateframework.github.io/specification/latest/index.html#fetch-target

### DIFF
--- a/sigstore-java/src/main/java/dev/sigstore/tuf/Updater.java
+++ b/sigstore-java/src/main/java/dev/sigstore/tuf/Updater.java
@@ -420,7 +420,10 @@ public class Updater {
         throw new FileNotFoundException(targetName, fetcher.getSource());
       }
       verifyHashes(entry.getKey(), targetBytes, targetData.getHashes());
-      localStore.storeTargetFile(versionedTargetName, targetBytes);
+
+      // when persisting targets use the targetname without sha512 prefix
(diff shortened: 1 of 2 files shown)


asraa (approved):
:partying_face:

sigstore-java/src/main/java/dev/sigstore/tuf/Updater.java
```diff
@@ -420,7 +420,10 @@ void downloadTargets(Targets targets)
         throw new FileNotFoundException(targetName, fetcher.getSource());
       }
       verifyHashes(entry.getKey(), targetBytes, targetData.getHashes());
-      localStore.storeTargetFile(versionedTargetName, targetBytes);
+
+      // when persisting targets use the targetname without sha512 prefix
```
vlsi, line 424: The link you mention suggests checksum-prefixed names for consistent snapshots scheme. Why we should skip checksums here?

asraa: This is the relevant line of the link:
```
In either case, the client MUST write the file to non-volatile storage as FILENAME.EXT.
```

The code already does fetch by the checksummed prefixed name (`versionedTargetName`), but we write without the prefix.

patflynn (approved):

PR #413, opened 2023-04-05
don't fail if fuzzOut isn't specified

### LINKED ISSUE #412: Build fails with Could not get unknown property 'fuzzOut' for project ':fuzzing
`./gradlew :tasks`

```
> Configure project :
Building Sigstore Java 0.4.0-SNAPSHOT

> Task :tasks FAILED

Build sigstore-java-root FAILURE reason:
    Execution failed for task ':tasks':
        org.gradle.api.internal.tasks.DefaultTaskContainer$TaskCreationException: Could not create task ':fuzzing:copyToFuzzOut'.
            at org.gradle.api.internal.tasks.DefaultTaskContainer.taskCreationException(DefaultTaskContainer.java:715)
            at org.gradle.api.internal.tasks.DefaultTaskContainer.access$600(DefaultTaskContainer.java:76)
            at org.gradle.api.internal.tasks.DefaultTaskContainer$TaskCreatingProvider.domainObjectCreationException(DefaultTaskContainer.java:707)
...
            Caused by: groovy.lang.MissingPropertyException: Could not get unknown property 'fuzzOut' for project ':fuzzing' of type org.gradle.api.Project.
                at org.gradle.internal.metaobject.AbstractDynamicObject.getMissingProperty(AbstractDynamicObject.java:85)
                at org.gradle.internal.metaobject.AbstractDynamicObject.getProperty(AbstractDynamicObject.java:62)
                at org.gradle.api.internal.project.DefaultProject.property(DefaultProject.java:1097)
                at Build_gradle$3.invoke(build.gradle.kts:16)
                at Build_gradle$3.invoke(build.gradle.kts:1)
                at org.gradle.api.internal.DefaultMutationGuard$2.execute(DefaultMutationGuard.java:44)
                at org.gradle.api.internal.DefaultMutationGuard$2.execute(DefaultMutationGuard.java:44)

```

loosebazooka:
fixes #412

### DIFF
--- a/fuzzing/build.gradle.kts
+++ b/fuzzing/build.gradle.kts
@@ -11,7 +11,10 @@ dependencies {
     implementation("com.code-intelligence:jazzer-api:0.16.0")
 }
 
+// copy to the fuzzing builder's output directory. This is an existing directory with
+// files in it, so don't use sync
 tasks.register<Copy>("copyToFuzzOut") {
     dependsOn(tasks.build)
-    into(project.property("fuzzOut")).from(sourceSets.main.get().runtimeClasspath)
+    into(project.findProperty("fuzzOut") ?: project.layout.buildDirectory.dir("fuzzOut"))
+    from(sourceSets.main.get().runtimeClasspath)
 }


fuzzing/build.gradle.kts
```diff
@@ -13,5 +13,7 @@ dependencies {
 
 tasks.register<Copy>("copyToFuzzOut") {
```
vlsi, line 14: The task should be `Sync` rather than `Copy` to remove stale files if some of the classes are removed from build to build

fuzzing/build.gradle.kts
```diff
@@ -13,5 +13,7 @@ dependencies {
...
     dependsOn(tasks.build)
-    into(project.property("fuzzOut")).from(sourceSets.main.get().runtimeClasspath)
+    val defaultDestination = project.layout.buildDirectory.dir("fuzzOut")
+    val dest = if (project.hasProperty("fuzzOut")) project.property("fuzzOut") ?: defaultDestination else defaultDestination
```
vlsi, line 17:
```suggestion
    val dest = project.findProperty("fuzzOut") ?: defaultDestination
```

fuzzing/build.gradle.kts
```diff
@@ -13,5 +13,7 @@ dependencies {
...
+    into(dest).from(sourceSets.main.get().runtimeClasspath)
```
vlsi, line 18:
```suggestion
    into(dest)
    from(sourceSets.main.get().runtimeClasspath)
```

loosebazooka: is this more idiomatic?

vlsi: I think so
They describe two different things:
1) where to store the data
2) which data to pick


For instance, if you would ever add the second `from`, then `into(...).from(..).from(..)` would look too long, while the following is fine:

```
into(...)
from(...)
from(...)
```

vlsi: Consider this:

```kotlin
    into(dest)
    into("abcd") {
        from(sourceSets.main.get().runtimeClasspath)
    }
```

=>

```
% tree build/fuzzOut
build/fuzzOut
└── abcd
    ├── animal-sniffer-annotations-1.21.jar
    ├── annotations-4.1.1.4.jar
    ├── bcpkix-jdk18on-1.72.jar
    ├── bcprov-jdk18on-1.72.jar
```

vlsi (dismissed):

loosebazooka:
ah we can't use Sync because the fuzzer builder is putting other things in the directory before gradle executes

vlsi:
Copying is inherently unsafe (sometime later one might end up with multiple bc-java jars having slightly different versions), so I would prefer:
a) explicitly mention which files should be preserved by sync. See https://github.com/apache/jmeter/blob/50e2015613752fbd0ceedc84440d0bea27848b91/src/dist/build.gradle.kts#L247-L255
b) document the reasons of having `copy` here

loosebazooka:
I mean (a) seems a bit much. Added in some comments.

vlsi (approved):

PR #415, opened 2023-04-05
v1 tuf client

loosebazooka:
Continuation of patricks work on this.

This is the sigstore wrapper around the Updater. It's very barebones right now, but it should allow us to easily access key/cert material defined in the trusted_root.json easier. This should also implement an interface that is passed to all verifiers (not yet defined).

### DIFF
--- a/sigstore-java/src/main/java/dev/sigstore/tuf/SigstoreTufClient.java
+++ b/sigstore-java/src/main/java/dev/sigstore/tuf/SigstoreTufClient.java
@@ -1,45 +1,78 @@
+/*
+ * Copyright 2023 The Sigstore Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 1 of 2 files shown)


sigstore-java/src/main/java/dev/sigstore/tuf/SigstoreTufClient.java
```diff
@@ -0,0 +1,145 @@
...
+    Duration cacheValidity = Duration.ofDays(1);
+    Path tufCacheLocation =
+        Path.of(System.getProperty("user.home")).resolve(".sigstore-java").resolve("root");
+
+    URL remoteMirror;
+    Path trustedRoot;
+
+    public Builder usePublicGoodInstance() {
+      if (remoteMirror != null || trustedRoot != null) {
+        throw new IllegalStateException();
```
patflynn, line 68: maybe just a message here explaining why you're throwing.

sigstore-java/src/main/java/dev/sigstore/tuf/SigstoreTufClient.java
```diff
@@ -0,0 +1,145 @@
...
+      return this;
+    }
+
+    public Builder tufCacheLocation(Path location) {
+      this.tufCacheLocation = location;
+      return this;
+    }
+
+    public SigstoreTufClient build() throws IOException {
+      Preconditions.checkState(
```
patflynn, line 98: why not just accept 0 as no caching?

patflynn (approved):

patflynn (approved):

sigstore-java/src/main/java/dev/sigstore/tuf/SigstoreTufClient.java
loosebazooka: yeah, that works.

patflynn (approved):

PR #417, opened 2023-04-07
chore: set version to 0.4.2 for proper release of sigstore-sign Gradle plugin

vlsi:
#### Summary

We need to have this change so the default `sigstore.sign` Gradle plugin would work

### DIFF
--- a/sigstore-gradle/sigstore-gradle-sign-base-plugin/src/main/kotlin/dev/sigstore/sign/SigstoreSignExtension.kt
+++ b/sigstore-gradle/sigstore-gradle-sign-base-plugin/src/main/kotlin/dev/sigstore/sign/SigstoreSignExtension.kt
@@ -39,7 +39,7 @@ abstract class SigstoreSignExtension(private val project: Project) {
     abstract val sigstoreJavaVersion : Property<String>
 
     init {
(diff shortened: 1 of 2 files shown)


sigstore-gradle/sigstore-gradle-sign-base-plugin/src/main/kotlin/dev/sigstore/sign/SigstoreSignExtension.kt
```diff
@@ -39,7 +39,7 @@ abstract class SigstoreSignExtension(private val project: Project) {
     abstract val sigstoreJavaVersion : Property<String>
 
     init {
-        sigstoreJavaVersion.convention("0.5.0")
+        sigstoreJavaVersion.convention("0.4.2")
```
ljacomet, line 42: Unless we release `sigstore-java` as well on Maven Central, this should remain `0.4.0`

ljacomet (commented):
See question / comment below

ljacomet (approved):

ljacomet (approved):
Required changes for a good `0.4.2` release 👍 

I guess everything should move back to `0.5.0` on the `main` branch once that's in and the release of the Gradle plugin done.

PR #432, opened 2023-04-25
Add accessors to trustroot

loosebazooka:
We need to be able to query the trustroot for CAs and logs to initialize our signers and based on the material we are signing.

This will eventually allow us to init a client directly from a trustroot

### DIFF
--- a/sigstore-java/src/main/java/dev/sigstore/trustroot/SigstoreTrustedRoot.java
+++ b/sigstore-java/src/main/java/dev/sigstore/trustroot/SigstoreTrustedRoot.java
@@ -17,35 +17,42 @@ package dev.sigstore.trustroot;
 
 import dev.sigstore.proto.trustroot.v1.TrustedRoot;
 import java.security.cert.CertificateException;
-import java.util.ArrayList;
-import java.util.List;
-import java.util.stream.Collectors;
 import org.immutables.value.Value.Immutable;
 
 @Immutable
 public interface SigstoreTrustedRoot {
 
-  List<CertificateAuthority> getCertificateAuthorities();
+  /** A list of certificate authorities associated with this trustroot. */
+  CertificateAuthorities getCAs();
 
-  List<TransparencyLog> getTLogs();
+  /** A list of binary transparency logs associated with this trustroot. */
+  TransparencyLogs getTLogs();
 
-  List<TransparencyLog> getCTLogs();
+  /** A list of certificate transparency logs associated with this trustroot. */
+  TransparencyLogs getCTLogs();
 
+  /** Create an instance from a parsed proto definition of a trustroot. */
   static SigstoreTrustedRoot from(TrustedRoot proto) throws CertificateException {
-    List<CertificateAuthority> certificateAuthorities =
-        new ArrayList<>(proto.getCertificateAuthoritiesCount());
+    var certificateAuthoritiesBuilder = ImmutableCertificateAuthorities.builder();
     for (var certAuthority : proto.getCertificateAuthoritiesList()) {
-      certificateAuthorities.add(CertificateAuthority.from(certAuthority));
+      certificateAuthoritiesBuilder.addCertificateAuthority(
+          CertificateAuthority.from(certAuthority));
     }
-    var tlogs =
-        proto.getTlogsList().stream().map(TransparencyLog::from).collect(Collectors.toList());
-    var ctlogs =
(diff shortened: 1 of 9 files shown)


sigstore-java/src/test/java/dev/sigstore/trustroot/SigstoreTrustedRootTest.java
```diff
@@ -117,4 +132,155 @@ void from_prod() throws Exception {
...
         Base64.toBase64String(PublicKey.toJavaPublicKey(currCTLog.getPublicKey()).getEncoded()));
   }
+
+  @Test
+  public void getTlog_isPresent() {
+    var key = Base64.decode("wNI9atQGlz+VWfO6LRygH4QUfY/8W4RFwiT5i5WRgB0=");
+
+    // exactly at start time
+    assertIsPresent(
+        trustRoot.getTlog(key, ZonedDateTime.parse("2021-01-12T11:53:27.000Z").toInstant()));
```
vlsi, line 142: Have you considered factoring out `assertGetTLogExists` / `assertGetTLogMissing` so the assertion failures could include input data rather than "expected true got false"?

loosebazooka: I think the failure is pretty obvious from the data/test-line-number? In particular, someone would need to inspect the full trust root json and then decide why the test is failing.

It seems to me that changes would make the tests a bit more complicated that it needs to be? But maybe I'm not seeing this the same way you are?

sigstore-java/src/main/java/dev/sigstore/trustroot/SigstoreTrustedRoot.java
```diff
@@ -48,4 +56,114 @@ static SigstoreTrustedRoot from(TrustedRoot proto) throws CertificateException {
...
+        .findAny();
+  }
+
+  /**
+   * Get the one an only current TLog
+   *
+   * @return the current active TLog
+   * @throws IllegalStateException if trust root does not contain exactly one transparency log
+   */
+  public Optional<TransparencyLog> getCurrentTLog() {
```
patflynn, line 99: why is this optional then?

sigstore-java/src/main/java/dev/sigstore/trustroot/SigstoreTrustedRoot.java
```diff
@@ -48,4 +56,114 @@ static SigstoreTrustedRoot from(TrustedRoot proto) throws CertificateException {
...
+  public Optional<CertificateAuthority> getCurrentCA() {
+    return getCurrentCA(getCertificateAuthorities());
+  }
+
+  @VisibleForTesting
+  static Optional<CertificateAuthority> getCurrentCA(
+      List<CertificateAuthority> certificateAuthorities) {
+    var current =
+        certificateAuthorities.stream()
+            .filter(ca -> ca.getValidFor().getEnd().isEmpty())
```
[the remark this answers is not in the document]
loosebazooka: @znewman01 @kommendorkapten could it?

sigstore-java/src/main/java/dev/sigstore/trustroot/SigstoreTrustedRoot.java
loosebazooka: Yeah it shouldn't be optional. I guess I never updated the signatures.

sigstore-java/src/main/java/dev/sigstore/trustroot/SigstoreTrustedRoot.java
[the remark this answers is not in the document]
kommendorkapten: Technically they could, but the _intention_ is to have the current with an unspecified end. I think this is a good callout, I'll create a PR to clarify this: https://github.com/sigstore/protobuf-specs/pull/80.

If the current has a known end, and due to operational constraints there may be an overlap of the validity time for a CA, so if we need to understand what CA is the current we would need to guess based on the time range which can be error-prone.

sigstore-java/src/test/java/dev/sigstore/trustroot/SigstoreTrustedRootTest.java
loosebazooka: okay, I think coming back to this, I see what you're saying, lemme try it out

loosebazooka:
reworked this a bit to put the logic in `CertificateAuthorities` and `TranspencyLogs`. We can now pass just those objects when necessary instead of the whole trust root?

PR #433, opened 2023-04-26
Add fuzzers for FulcioVerifier

arthurscchan:
#### Summary

Add fuzzer for FulcioVerifier, FulcioCertificateVerifier and SigningCertificate class

### DIFF
--- a/fuzzing/src/main/java/fuzzing/FulcioCertificateVerifierFuzzer.java
+++ b/fuzzing/src/main/java/fuzzing/FulcioCertificateVerifierFuzzer.java
@@ -21,29 +21,33 @@ import dev.sigstore.fulcio.client.FulcioCertificateVerifier;
 import dev.sigstore.fulcio.client.FulcioVerificationException;
 import java.io.ByteArrayInputStream;
 import java.nio.charset.Charset;
-import java.security.cert.CertificateException;
 import java.security.cert.CertificateFactory;
 import java.security.cert.X509Certificate;
 import java.util.List;
 
 public class FulcioCertificateVerifierFuzzer {
   public static void fuzzerTestOneInput(FuzzedDataProvider data) {
+    byte[] byteArray = data.consumeRemainingAsBytes();
+    String string = new String(byteArray, Charset.defaultCharset());
+
+    X509Certificate certificate;
     try {
       CertificateFactory cf = CertificateFactory.getInstance("X.509");
+      certificate = (X509Certificate) cf.generateCertificate(new ByteArrayInputStream(byteArray));
+    } catch (Exception e) {
+      // Skipping this iteration if exceptions thrown during certificate creation
+      return;
+    }
 
-      byte[] byteArray = data.consumeRemainingAsBytes();
-      String string = new String(byteArray, Charset.defaultCharset());
-
+    try {
       FulcioCertificateVerifier verifier = new FulcioCertificateVerifier();
-      X509Certificate certificate =
-          (X509Certificate) cf.generateCertificate(new ByteArrayInputStream(byteArray));
       List list =
           List.of(
               CertificateIdentity.builder().subjectAlternativeName(string).issuer(string).build(),
               CertificateIdentity.builder().subjectAlternativeName(string).issuer(string).build());
 
       verifier.verifyCertificateMatches(certificate, list);
-    } catch (CertificateException | FulcioVerificationException e) {
+    } catch (FulcioVerificationException e) {
       // Known exception
     }
   }
--- a/fuzzing/src/main/java/fuzzing/FulcioCertificateVerifierFuzzer.java
+++ b/fuzzing/src/main/java/fuzzing/FulcioCertificateVerifierFuzzer.java
@@ -1,0 +1,50 @@
+/*
+ * Copyright 2023 The Sigstore Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


arthurscchan:
@AdamKorcz

fuzzing/src/main/java/fuzzing/FulcioVerifierFuzzer.java
```diff
@@ -0,0 +1,89 @@
...
+            break;
+          case 3:
+            fv.verifyCertChain(sc);
+            break;
+          case 4:
+            fcv.verifyCertificateMatches(sc.getLeafCertificate(), ciList);
+            break;
+        }
+      }
+    } catch (CertificateException
```
loosebazooka, line 80: Presumably you can catch all checked exceptions with catch (Exception e)... I don't know that listing them out like this really matters? But if everyone does it this way, that's cool.

fuzzing/src/main/java/fuzzing/FulcioVerifierFuzzer.java
```diff
@@ -0,0 +1,89 @@
...
+      ciList.add(
+          CertificateIdentity.builder().subjectAlternativeName(string).issuer(string).build());
+      certList.add(cf.generateCertificate(new ByteArrayInputStream(byteArray)));
+      certList.add(cf.generateCertificate(new ByteArrayInputStream(byteArray)));
+      byteArrayList.add(byteArray);
+      byteArrayList.add(byteArray);
+
+      SigningCertificate sc = SigningCertificate.from(cf.generateCertPath(certList));
+      FulcioVerifier fv = FulcioVerifier.newFulcioVerifier(byteArray, byteArrayList);
+      FulcioCertificateVerifier fcv = new FulcioCertificateVerifier();
```
loosebazooka, line 59: I think `fcv` can be fuzzed independently of `fv`, would help isolate the problem space a bit, wdyt?

fuzzing/src/main/java/fuzzing/FulcioVerifierFuzzer.java
AdamKorcz: I think it is worth it to start by granularly catching the exceptions that we _must_ catch to see if something unexpected comes up.

fuzzing/src/main/java/fuzzing/FulcioVerifierFuzzer.java
arthurscchan: Has separated them into 2 fuzzers.

AdamKorcz:
@arthurscchan Can you change the title (including in the git commit message) to indicate fuzzers in plural?

fuzzing/src/main/java/fuzzing/FulcioVerifierFuzzer.java
```diff
@@ -0,0 +1,76 @@
...
+          case 1:
+            sc.getLeafCertificate();
+            break;
+          case 2:
+            fv.verifySct(sc);
+            break;
+          case 3:
+            fv.verifyCertChain(sc);
+            break;
+        }
```
loosebazooka, line 65: This block of code seems a bit wonky, I can't seem to find a reason for this randomness? Why not just run through all the options in one go?
```suggestion
            sc.getCertificates();
            sc.getLeafCertificate();
            fv.verifySct(sc);
            fv.verifyCertChain(sc);
```

AdamKorcz: The motivation for randomizing the method calls is to see if any order triggers any issues. For example, this way the fuzzer will test all combinations of the method calls instead of a single one.

loosebazooka: oh so assuming that they change the state of the objects? I hope this doesn't affect us, but I guess that makes sense. Can we document this for future people looking at this?

fuzzing/src/main/java/fuzzing/FulcioCertificateVerifierFuzzer.java
```diff
@@ -0,0 +1,50 @@
...
+public class FulcioCertificateVerifierFuzzer {
+  public static void fuzzerTestOneInput(FuzzedDataProvider data) {
+    try {
+      CertificateFactory cf = CertificateFactory.getInstance("X.509");
+
+      byte[] byteArray = data.consumeRemainingAsBytes();
+      String string = new String(byteArray, Charset.defaultCharset());
+
+      FulcioCertificateVerifier verifier = new FulcioCertificateVerifier();
+      X509Certificate certificate =
```
loosebazooka, line 38: I would prefer not to run into fuzzing issues with X509Certificate generation, can we short circuit this?
```
try {
  cf.generateCertificate(new ByteArrayInputStream(byteArray));
} catch (Exception | RuntimeException e) {
  // end fuzzing run because we don't have a valid certificate?
  return;
}
```

arthurscchan: Have separate them. Thanks for the suggestions.

fuzzing/src/main/java/fuzzing/FulcioVerifierFuzzer.java
loosebazooka: Oh right, someone I forgot RuntimeException was included in Exception. :shrug:

PR #434, opened 2023-04-26
Description of trusted_root.json usage

loosebazooka:
wip on plumbing the trusted_root.json through sigstore-java

### DIFF
--- a/docs/trustroot.md
+++ b/docs/trustroot.md
@@ -1,0 +1,42 @@
+# Plumbing the sigstore trustroot through sigstore-java
+
+https://github.com/sigstore/root-signing/blob/main/targets/trusted_root.json
+
+The trust root should be the mechanism by which we initialize all our clients.
+
+## Assumptions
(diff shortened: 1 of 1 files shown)


loosebazooka:
@vlsi lemme know if this makes sense
@znewman01 is this closer to what you had in mind?

znewman01 (commented):
Yup, three points:

First, it might be worth noting that you *don't* want to use intermediate certs for verification that don't come from the trust root.

Second:
> Verication can occur against verification material from the entire history of sigstore (unless there's some deprecation period - there isn't yet?)

Yeah this is the plan. Right now the cert validity periods themselves are what we're using, eventually we'll have extra metadata in TUF limiting the validity window even more.


Third, better docs in the trust root repo would be good so that not everybody has to rewrite this :)

loosebazooka:
Well this stuff is implemented. The actually description of trust root should be in: github.com/sigstore/protobuf-specs

PR #436, opened 2023-04-27
Bump `sigstore-conformance` to 0.0.4

tetsuo-cpp:
There's been a few changes since the last release.

We've figured out a new scheme for getting OIDC tokens. This means:

- We no longer need to give third-party PRs access to secrets via `pull_request_target`.
- We no longer need a "safe to test" label to protect against third-party PRs maliciously using secrets.

The way it works is that we have a separate PR that is separated from the Sigstore org that is responsible for continually generating OIDC tokens. The conformance suite will grab the latest one and then use it to run the test suite. During times of heavy GitHub Actions traffic, there can be a bit of a wait before a new token is generated however, you only really need to see this pass once before merging a PR so I think this is acceptable.

In order to get this to work, I've added an `--identity-token` flag to the `sigstore-cli` similar to what we have in `sigstore-python`, `cosign`, etc.

### DIFF
--- a/.github/workflows/conformance.yml
+++ b/.github/workflows/conformance.yml
@@ -32,6 +32,6 @@ jobs:
       - name: Unpack sigstore-java distribution
         run: tar -xvf ${{ github.workspace }}/sigstore-cli/build/distributions/sigstore-cli-*.tar --strip-components 1
 
-      - uses: trailofbits/sigstore-conformance@064fb32a890c30235f305281f3509c5e65e6f9e5 # v0.0.4
+      - uses: sigstore/sigstore-conformance@064fb32a890c30235f305281f3509c5e65e6f9e5 # tag=v0.0.4
         with:
           entrypoint: ${{ github.workspace }}/bin/sigstore-cli
--- a/.github/workflows/conformance.yml
+++ b/.github/workflows/conformance.yml
@@ -14,9 +14,6 @@ jobs:
         java-version: [11, 17]
       fail-fast: false
 
-    permissions:
-      # Needed to access the workflow's OIDC identity.
-      id-token: write
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@8e5e7e5ab8b370d6c329ec480221332ada57f0ab # v3.5.2
--- a/.github/workflows/conformance.yml
+++ b/.github/workflows/conformance.yml
@@ -4,8 +4,7 @@ on:
   push:
     branches:
       - main
-  pull_request_target:
-    types: [labeled]
+  pull_request:
   workflow_dispatch:
 
 jobs:
--- a/.github/workflows/conformance.yml
+++ b/.github/workflows/conformance.yml
@@ -3,9 +3,8 @@ name: Conformance Tests
 on:
   push:
     branches:
-      - main
+      - '**'
   pull_request:
-  workflow_dispatch:
 
 jobs:
   conformance:
--- a/sigstore-java/src/main/java/dev/sigstore/oidc/client/TokenStringOidcClient.java
+++ b/sigstore-java/src/main/java/dev/sigstore/oidc/client/TokenStringOidcClient.java
@@ -1,0 +1,47 @@
+/*
+ * Copyright 2023 The Sigstore Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package dev.sigstore.oidc.client;
+
+import com.google.api.client.json.gson.GsonFactory;
+import com.google.api.client.json.webtoken.JsonWebSignature;
+import java.io.IOException;
+
+public class TokenStringOidcClient implements OidcClient {
+
+  private final String idToken;
+
+  public TokenStringOidcClient(String idToken) {
+    this.idToken = idToken;
+  }
+
+  @Override
+  public boolean isEnabled() {
+    return true;
+  }
+
+  @Override
+  public OidcToken getIDToken() throws OidcException {
+    try {
+      var jws = JsonWebSignature.parse(new GsonFactory(), idToken);
+      return ImmutableOidcToken.builder()
+          .idToken(idToken)
+          .subjectAlternativeName(jws.getPayload().getSubject())
+          .build();
+    } catch (IOException e) {
+      throw new OidcException("Failed to parse JWT", e);
+    }
+  }
+}
--- a/sigstore-cli/build.gradle.kts
+++ b/sigstore-cli/build.gradle.kts
@@ -12,6 +12,9 @@ dependencies {
     implementation("info.picocli:picocli:4.7.3")
     implementation("com.google.guava:guava:31.1-jre")
 
+    implementation(platform("com.google.oauth-client:google-oauth-client-bom:1.34.1"))
+    implementation("com.google.oauth-client:google-oauth-client")
+
     annotationProcessor("info.picocli:picocli-codegen:4.7.3")
 }
 
--- a/sigstore-cli/src/main/java/dev/sigstore/cli/Sign.java
+++ b/sigstore-cli/src/main/java/dev/sigstore/cli/Sign.java
@@ -18,12 +18,15 @@ package dev.sigstore.cli;
 import dev.sigstore.KeylessSigner;
 import dev.sigstore.bundle.BundleFactory;
 import dev.sigstore.encryption.certificates.Certificates;
+import dev.sigstore.oidc.client.OidcClients;
+import dev.sigstore.oidc.client.TokenStringOidcClient;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Path;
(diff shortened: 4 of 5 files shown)


sigstore-java/src/main/java/dev/sigstore/oidc/client/TokenStringOidcClient.java
```diff
@@ -0,0 +1,47 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package dev.sigstore.oidc.client;
+
+import com.google.api.client.json.gson.GsonFactory;
+import com.google.api.client.json.webtoken.JsonWebSignature;
+import java.io.IOException;
+
+public class TokenStringOidcClient implements OidcClient {
```
tetsuo-cpp, line 22: I needed an `OidcClient` to just give me the an `OidcToken` for a given token string.

vlsi:
Would you please remove `id-token: write` from the workflow definition then?

vlsi:
How do we get the token for the unit tests?

tetsuo-cpp:
> How do we get the token for the unit tests?

It grabs the token from [this repo](https://github.com/sigstore-conformance/extremely-dangerous-public-oidc-beacon). That repo continually generates OIDC tokens so the unit tests will just grab the latest one and use it.

vlsi:
>That repo continually generates OIDC tokens so the unit tests will just grab the latest one and use it.

Suppose we have a unit test in Java. How does the test retrieve the token?

For instance, we retrieve regular GitHub token by making a HTTP call to `ACTIONS_ID_TOKEN_REQUEST_URL`.
https://github.com/sigstore/sigstore-java/blob/9bebb37716c5b7f7d5917764e2103e0fece11425/sigstore-java/src/main/java/dev/sigstore/oidc/client/GithubActionsOidcClient.java#L88

I would expect there should be a similar discovery approach for `sigstore-conformance`-generated tokens.
For instance, it could expose the token via environment variables, then we could pick up the token from there.

Even though having an explicit `--identity-token` parameter might help for testing (e.g. to ensure which token is used), however, we'll need a discovery-based approach as well.

tetsuo-cpp:
> Suppose we have a unit test in Java. How does the test retrieve the token?

At the moment, the conformance suite retrieves the token and passes it into the client via `--identity-token`. It sounds like you're talking about having a custom `OidcClient` implementation that grabs the conformance suite token.

What use case do you have in mind? We didn't really design for this to be used more generally.

tetsuo-cpp:
> Apparently, the test requires OIDC token, so currently we guard the test with `safe to test` label.

Hmm, where is that guard? I'm grepping for "safe to test" and I can't find it.

Anyway, that's interesting. I guess we could technically separate out the code that retrieves the token and create some kind of script that exports it as an environment variable. But the logic isn't all that complicated anyway: https://github.com/sigstore/sigstore-conformance/blob/main/action.py#L40.

For your purpose, it might make more sense to just implement an `OidcClient` that does what that Python function does.

sigstore-java/src/main/java/dev/sigstore/oidc/client/TokenStringOidcClient.java
```diff
@@ -0,0 +1,47 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package dev.sigstore.oidc.client;
+
+import com.google.api.client.json.gson.GsonFactory;
+import com.google.api.client.json.webtoken.JsonWebSignature;
+import java.io.IOException;
+
```
loosebazooka, line 21: A comment up here would be nice, perhaps just describing what the string *should* look like, even if it's just a link. If use of this is limited to just conformance, then maybe it should go in the sigstore-cli project instead?

.github/workflows/conformance.yml
```diff
@@ -4,8 +4,7 @@ on:
   push:
     branches:
       - main
-  pull_request_target:
-    types: [labeled]
+  pull_request:
```
loosebazooka, line 7: If this is not a particularly expensive operation, I would like the config to match our other CI?
```
on:
  push:
    branches:
      - '*'
  pull_request:
    branches:
      - main
```

loosebazooka (commented):
I think this actually looks fine for the purposes of conformance testing. The conformance testing infra will inject the token as necessary and best of all we get rid of the "safe-to-test" label :tada: and allow this run on the code the PR is.

@vlsi, if suppose if we want to add conformance testing to the gradle plugin, then we'll can create the appropriate shim. But I don't know that we want to deal with that in this PR. `safe-to-test` was only a gate for conformance testing.

.github/workflows/conformance.yml
```diff
@@ -36,6 +32,6 @@ jobs:
       - name: Unpack sigstore-java distribution
         run: tar -xvf ${{ github.workspace }}/sigstore-cli/build/distributions/sigstore-cli-*.tar --strip-components 1
 
-      - uses: trailofbits/sigstore-conformance@0748d63c53810e36cc3f4bbe4114301080f0d844 # tag=v0.0.3
+      - uses: trailofbits/sigstore-conformance@064fb32a890c30235f305281f3509c5e65e6f9e5 # tag=v0.0.4
```
woodruffw, line 35: This is using a stale alias:

```suggestion
      - uses: sigstore/sigstore-conformance@064fb32a890c30235f305281f3509c5e65e6f9e5 # tag=v0.0.4
```

sigstore-java/src/main/java/dev/sigstore/oidc/client/TokenStringOidcClient.java
woodruffw: Unless I'm misunderstanding, the string in question is the OIDC token, which means that it's shaped like an ordinary JWT. Happy to add a comment to that effect, if you think it'd be useful!

vlsi:
> safe-to-test was only a gate for conformance testing

That is right. I thought we guarded all `id-token:write` with safe-to-test label.
Apparently we test Gradle plugin only when running CI from the main branch :-/

.github/workflows/conformance.yml
loosebazooka: yeah that would be great.

.github/workflows/conformance.yml
```diff
@@ -4,8 +4,7 @@ on:
   push:
     branches:
       - main
-  pull_request_target:
-    types: [labeled]
+  pull_request:
   workflow_dispatch:
```
loosebazooka, line 8: I don't think we necessarily *need* workflow dispatch anymore either?

sigstore-java/src/main/java/dev/sigstore/oidc/client/TokenStringOidcClient.java
loosebazooka: uh it's just that it's in the sigstore-java library and hence is available through it's api, so we want to document it. I think it should just go into the sigstore-cli project for now though. I don't see when we would be using this.

loosebazooka:
pretty cool that conformance is passing :D without safe-to-test

sigstore-java/src/main/java/dev/sigstore/oidc/client/TokenStringOidcClient.java
vlsi: `TokenStringOidcClient` provokes use of static token (which is fine for testing purposes), however, in practice the tokes will be short-lived, so I doubt static `TokenStringOidcClient` would be useful for sigstore-java users.

It might be a good idea to move `TokenStringOidcClient` to `sigstore-cli` module.

.github/workflows/conformance.yml
woodruffw: Nope, shouldn't be necessary. You can keep that around if you'd like to be able to manually trigger conformance tests, but IIRC we don't both with that in the `sigstore-python` tests either.

.github/workflows/conformance.yml
```diff
@@ -4,8 +4,7 @@ on:
   push:
     branches:
       - main
```
woodruffw, line 6:
```suggestion
      - '**'
```

.github/workflows/conformance.yml
woodruffw, line 8:
```suggestion
```

.github/workflows/conformance.yml
woodruffw: Done! I made the pattern `**`, since a single `*` will only match branches that don't contain a `/`.

loosebazooka:
hey lgtm, can we just move over the TokenStringOidcClient over to the sigstore-cli project?

`sigstore-java/src/main/java/dev/sigstore/oidc/client/TokenStringOidcClient.java` 
-> `sigstore-cli/src/main/java/dev/sigstore/cli/oidc/TokenStringOidcClient.java`

tetsuo-cpp:
> hey lgtm, can we just move over the TokenStringOidcClient over to the sigstore-cli project?
> 
> `sigstore-java/src/main/java/dev/sigstore/oidc/client/TokenStringOidcClient.java` -> `sigstore-cli/src/main/java/dev/sigstore/cli/oidc/TokenStringOidcClient.java`

Yep! Will do that.

sigstore-java/src/main/java/dev/sigstore/oidc/client/TokenStringOidcClient.java
tetsuo-cpp: [55da240](https://github.com/sigstore/sigstore-java/pull/436/commits/55da240337cff33c74af610607bf379048522df7)

loosebazooka (dismissed):

loosebazooka:
Sorry @tetsuo-cpp can you rebase this

tetsuo-cpp:
> Sorry @tetsuo-cpp can you rebase this

No worries! Will do that tonight.

loosebazooka (approved):

PR #438, opened 2023-04-28
Add fuzzer for RekorVerifier

arthurscchan:
#### Summary

Add fuzzer for dev.sigstore.rekor.client.RekorVerifier class.

### DIFF
--- a/fuzzing/src/main/java/fuzzing/RekorVerifierFuzzer.java
+++ b/fuzzing/src/main/java/fuzzing/RekorVerifierFuzzer.java
@@ -28,7 +28,7 @@ import java.security.NoSuchAlgorithmException;
 import java.security.spec.InvalidKeySpecException;
 
 public class RekorVerifierFuzzer {
-  private final static String URL = "https://false.url.for.RekorTypes.fuzzing.com";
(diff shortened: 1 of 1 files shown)


arthurscchan:
@AdamKorcz

fuzzing/src/main/java/fuzzing/RekorVerifierFuzzer.java
```diff
@@ -0,0 +1,56 @@
...
+
+      URI uri = new URI(URL);
+      RekorEntry entry = RekorResponse.newRekorResponse(uri, string).getEntry();
+      RekorVerifier verifier = RekorVerifier.newRekorVerifier(byteArray);
+
+      if (choice) {
+        verifier.verifyEntry(entry);
+      } else {
+        verifier.verifyInclusionProof(entry);
+      }
```
loosebazooka, line 47: Yeah so I'm not fully convinced this type of choice based workflow makes sense. In my mind they should just be separate fuzz tests or just run one after the other?

arthurscchan: Change accordingly

loosebazooka (dismissed):

loosebazooka (approved):

PR #455, opened 2023-05-24
Fix key fuzzer for more coverage

arthurscchan:
#### Summary

Currently, the Keys.constructTufPublicKey method will not be called if Keys.parsePublicKey throws an exception. This PR changes the logic of KeysFuzzer by separating the two method call into two separate fuzzers.

### DIFF
--- a/fuzzing/src/main/java/fuzzing/KeysFuzzer.java
+++ b/fuzzing/src/main/java/fuzzing/KeysFuzzer.java
@@ -36,6 +36,10 @@ public class KeysFuzzer {
       }
     } catch (IOException | InvalidKeySpecException | NoSuchAlgorithmException e) {
       // known exceptions
+    } catch (RuntimeException e) {
+      if (!e.toString().contains("not currently supported")) {
+        throw e;
+      }
     }
   }
 }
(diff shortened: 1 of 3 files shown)


arthurscchan:
@AdamKorcz

fuzzing/src/main/java/fuzzing/KeysFuzzer.java
```diff
@@ -26,12 +26,20 @@ public static void fuzzerTestOneInput(FuzzedDataProvider data) {
     try {
       String[] schemes = {"rsassa-pss-sha256", "ed25519", "ecdsa-sha2-nistp256"};
       String scheme = data.pickValue(schemes);
+      Boolean choice = data.consumeBoolean();
       byte[] byteArray = data.consumeRemainingAsBytes();
 
-      Keys.parsePublicKey(byteArray);
-      Keys.constructTufPublicKey(byteArray, scheme);
+      if (choice) {
```
vlsi, line 32: What do you think of splitting in into several fuzzers?

arthurscchan: Thanks for the suggestion.

loosebazooka:
replaced by #626

PR #469, opened 2023-07-31
use base google-http-client-bom

hboutemy:
#### Summary
Using `com.google.cloud:libraries-bom` imports many other nested BOMs, even if they are not used at the end: the fact is that it confuses Maven dependencyManagement import on generated pom.xml
Going straight to `com.google.http-client:google-http-client-bom` is just more efficient and clear


vlsi (approved):
Using `com.google.http-client` for `com.google.http-client` dependencies improves readability indeed

PR #472, opened 2023-08-02
test: display problematic rawResponse in newRekorResponse failures

vlsi:
#### Summary

We have `RekorParseException`, and it seems whe should wrap `JsonParseException` with our own `RekorParseException`.
It looks like Gson is more like implementation details, so we should not expose `JsonParseException` in our APIs.

#### Release Note

NONE

#### Documentation

NONE


loosebazooka:
I think using `@Lazy` is fine for now.

vlsi:
`@Derived` might be slightly better as we would be able to reject an "invalid" message earlier, so we don't accidentally accept is as a valid one.

loosebazooka:
close this?

vlsi:
Closing as it is not a priority for me now

PR #473, opened 2023-08-03
Fix fuzzing issues

loosebazooka:
- handle invalid json for rekor response
- handle bundle with no tlog entry

### DIFF
--- a/sigstore-java/src/main/java/dev/sigstore/rekor/client/RekorResponse.java
+++ b/sigstore-java/src/main/java/dev/sigstore/rekor/client/RekorResponse.java
@@ -18,6 +18,7 @@ package dev.sigstore.rekor.client;
 import static dev.sigstore.json.GsonSupplier.GSON;
 
 import com.google.common.reflect.TypeToken;
+import com.google.gson.JsonSyntaxException;
 import java.net.URI;
 import java.util.Map;
 import org.immutables.value.Value;
@@ -53,10 +54,15 @@ public interface RekorResponse {
    * @return an immutable {@link RekorResponse} instance
    * @throws RekorParseException if the rawResponse doesn't parse directly to a single rekor entry
    */
-  static RekorResponse newRekorResponse(URI entryLocation, String rawResponse)
+  public static RekorResponse newRekorResponse(URI entryLocation, String rawResponse)
       throws RekorParseException {
     var type = new TypeToken<Map<String, RekorEntry>>() {}.getType();
-    Map<String, RekorEntry> entryMap = GSON.get().fromJson(rawResponse, type);
+    Map<String, RekorEntry> entryMap;
+    try {
+      entryMap = GSON.get().fromJson(rawResponse, type);
+    } catch (JsonSyntaxException | NullPointerException | StringIndexOutOfBoundsException ex) {
+      throw new RekorParseException("Rekor entry json could not be parsed: " + rawResponse, ex);
+    }
     if (entryMap == null) {
       throw new RekorParseException("Expecting a single rekor entry in response but found none");
     }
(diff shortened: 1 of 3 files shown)


sigstore-java/src/main/java/dev/sigstore/rekor/client/RekorResponse.java
```diff
@@ -56,7 +57,12 @@ public interface RekorResponse {
   static RekorResponse newRekorResponse(URI entryLocation, String rawResponse)
       throws RekorParseException {
     var type = new TypeToken<Map<String, RekorEntry>>() {}.getType();
-    Map<String, RekorEntry> entryMap = GSON.get().fromJson(rawResponse, type);
+    Map<String, RekorEntry> entryMap;
+    try {
+      entryMap = GSON.get().fromJson(rawResponse, type);
+    } catch (JsonSyntaxException jse) {
+      throw new RekorParseException("Rekor entry json could not be parsed", jse);
```
vlsi, line 64: WDYT of mentioning the problematic response in the error message?

vlsi (dismissed):

sigstore-java/src/main/java/dev/sigstore/rekor/client/RekorResponse.java
loosebazooka: I think the stack trace should show it? but yeah I can just do jse.getMessage() too. I'll add that in.

loosebazooka:
turns out parsing bad data can create npes too.

loosebazooka:
I think we can logger.error the whole rekor response? Is that what you were kinda looking for?

vlsi (changes requested):

sigstore-java/src/main/java/dev/sigstore/rekor/client/RekorResponse.java
```diff
@@ -53,10 +56,17 @@ public interface RekorResponse {
...
       throws RekorParseException {
     var type = new TypeToken<Map<String, RekorEntry>>() {}.getType();
-    Map<String, RekorEntry> entryMap = GSON.get().fromJson(rawResponse, type);
+    Map<String, RekorEntry> entryMap;
+    try {
+      entryMap = GSON.get().fromJson(rawResponse, type);
+    } catch (JsonSyntaxException | NullPointerException | StringIndexOutOfBoundsException ex) {
+      log.severe("Rekor entry could not be parsed");
+      log.severe(rawResponse);
+      throw new RekorParseException("Rekor entry json could not be parsed: " + ex.getMessage(), ex);
```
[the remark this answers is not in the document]
loosebazooka: sgtm... I image that response could be pretty huge, but whatever, this appears to be a pretty edge error case.

sigstore-java/src/main/java/dev/sigstore/rekor/client/RekorResponse.java
```diff
@@ -53,10 +54,15 @@ public interface RekorResponse {
...
    */
-  static RekorResponse newRekorResponse(URI entryLocation, String rawResponse)
+  public static RekorResponse newRekorResponse(URI entryLocation, String rawResponse)
       throws RekorParseException {
     var type = new TypeToken<Map<String, RekorEntry>>() {}.getType();
-    Map<String, RekorEntry> entryMap = GSON.get().fromJson(rawResponse, type);
+    Map<String, RekorEntry> entryMap;
+    try {
+      entryMap = GSON.get().fromJson(rawResponse, type);
+    } catch (JsonSyntaxException | NullPointerException | StringIndexOutOfBoundsException ex) {
```
vlsi, line 63: we should report npe to gson team

vlsi (approved):

loosebazooka:
I think fuzzing caught a number format exception issue, but I'll fix that in a followup.

PR #477, opened 2023-08-18
Updates before applying tuf to fulcio client

loosebazooka:
- iterate over collection types
- allow appending to certpaths
- managed channels from string targets
- create trust roots directly on CA object
- private key conversion

### DIFF
--- a/sigstore-java/src/main/java/dev/sigstore/encryption/certificates/Certificates.java
+++ b/sigstore-java/src/main/java/dev/sigstore/encryption/certificates/Certificates.java
@@ -16,6 +16,7 @@
 package dev.sigstore.encryption.certificates;
 
 import com.google.api.client.util.PemReader;
+import com.google.common.collect.ImmutableList;
 import java.io.ByteArrayInputStream;
 import java.io.IOException;
 import java.io.StringReader;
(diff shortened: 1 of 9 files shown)


sigstore-java/src/main/java/dev/sigstore/encryption/certificates/Certificates.java
```diff
@@ -135,4 +136,16 @@ public static CertPath toCertPath(Certificate certificate) throws CertificateExc
     CertificateFactory cf = CertificateFactory.getInstance("X.509");
     return cf.generateCertPath(Collections.singletonList(certificate));
   }
+
+  /** Appends an X509Certificate to a {@link CertPath} as a leaf. */
+  public static CertPath appendCertPath(CertPath root, Certificate certificate)
```
loosebazooka, line 141: this is so we can recreate cert paths from provided certificates

sigstore-java/src/main/java/dev/sigstore/http/GrpcChannels.java
```diff
@@ -39,4 +39,23 @@ public static ManagedChannel newManagedChannel(URI serverUrl, HttpParams httpPar
...
+
+  /**
+   * Create a new managed channel, this may be reused across multiple requests to a host, and must
+   * be closed when finished.
+   *
+   * @param serverUrl the host to connect to
+   * @param httpParams the http configuration
+   * @return a reusable grpc channel
+   */
+  public static ManagedChannel newManagedChannel(String serverUrl, HttpParams httpParams) {
```
loosebazooka, line 51: when we extract the "authority" from a uri (Strip the "http...") it returns a string, so we're going to greate managed channels with a string.

vlsi (approved):

PR #478, opened 2023-08-18
configure fulcio (v2 for now) with trustroot

loosebazooka:
Create v2 of various fulcio utilities so we can plumb the tuf data through. These will eventually replace the original implementations, but we are not ready to do that yet.

Note: v2 fulcio client drops support for detached scts.

### DIFF
--- a/sigstore-java/src/main/java/dev/sigstore/fulcio/client/FulcioClient2.java
+++ b/sigstore-java/src/main/java/dev/sigstore/fulcio/client/FulcioClient2.java
@@ -1,0 +1,127 @@
+/*
+ * Copyright 2022 The Sigstore Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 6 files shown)


sigstore-java/src/main/java/dev/sigstore/fulcio/client/FulcioClient2.java
```diff
@@ -0,0 +1,129 @@
...
+      var certs =
+          client
+              .withDeadlineAfter(httpParams.getTimeout(), TimeUnit.SECONDS)
+              .createSigningCertificate(req);
+
+      if (certs.getCertificateCase() == SIGNED_CERTIFICATE_DETACHED_SCT) {
+        throw new CertificateException("Detached SCTs are not supported");
+      } else {
+        return SigningCertificate.newSigningCertificate(certs.getSignedCertificateEmbeddedSct());
+      }
```
vlsi, line 123: `else` is not needed after `throw` on the first `if` branch.

vlsi (dismissed):

vlsi (approved):

PR #512, opened 2023-09-11
update sigstore-conformance tests

loosebazooka:
add bundle specific verification options


loosebazooka:
nvm, this needs the updated verifier that rebuilds the certs.

PR #521, opened 2023-09-14
Update conformance tests

loosebazooka:
Failure occurred where the local SET is ignored when a remote SET is obtained.  Changes were made to ensure local SET and remote match up.


loosebazooka:
hrmm... okay so I didn't think this behavior out properly. Does online mode care if the bundle is partially broken?

vlsi (approved):

PR #546, opened 2023-10-18
Allow signers to specify allow list of oidc ids

loosebazooka:
- Also signer now throws KeylessSignerException


loosebazooka:
fixes #490

patflynn (approved):

PR #568, opened 2023-11-01
Rework keys to remove PemReader, user BC instead

loosebazooka:
PEM reader will error in unexpected ways because it is too permissive or poorly formed objects.

While I was reworking that, it made sense to just clean up the Keys stuff


loosebazooka:
this should fix some of the fuzzing issues in: https://bugs.chromium.org/p/oss-fuzz/issues/list?q=label:Proj-sigstore-java

sigstore-java/src/main/java/dev/sigstore/encryption/Keys.java
```diff
@@ -15,36 +15,38 @@
...
+import org.bouncycastle.openssl.jcajce.JcaPEMKeyConverter;
 import org.bouncycastle.util.encoders.DecoderException;
-import org.bouncycastle.util.io.pem.PemObject;
-import org.bouncycastle.util.io.pem.PemReader;
 
 /** For internal use. Key related utility functions. */
 public class Keys {
 
-  private static final Logger log = Logger.getLogger(Keys.class.getName());
+  // private static final Logger log = Logger.getLogger(Keys.class.getName());
```
patflynn, line 49: do you mean to keep this commented out?

sigstore-java/src/main/java/dev/sigstore/encryption/Keys.java
```diff
@@ -60,48 +62,26 @@ public class Keys {
...
+            new InputStreamReader(new ByteArrayInputStream(keyBytes), StandardCharsets.UTF_8))) {
+      var keyObj = pemParser.readObject(); // throws DecoderException
+      if (keyObj == null) {
         throw new InvalidKeySpecException(
             "sigstore public keys must be only a single PEM encoded public key");
       }
+      JcaPEMKeyConverter converter = new JcaPEMKeyConverter();
+      if (keyObj instanceof SubjectPublicKeyInfo) {
+        PublicKey pk = converter.getPublicKey((SubjectPublicKeyInfo) keyObj);
+        if (!List.of("ECDSA", "EC", "RSA", "Ed25519", "EdDSA").contains(pk.getAlgorithm())) {
```
patflynn, line 76: do we not have a static set of these defined somewhere?

sigstore-java/src/main/java/dev/sigstore/encryption/Keys.java
loosebazooka: Oh that just wasn't used. I'll remove it

sigstore-java/src/main/java/dev/sigstore/encryption/Keys.java
loosebazooka: Yeah I can make that a bit more reusable

patflynn (approved):

PR #599, opened 2023-12-27
Rework trusted root injection for conformance v0.0.9

loosebazooka:
While this will allow anyone to inject a trustedRoot instead of using sigstore public good, its been primarily included to allow conformance to inject alt trusted roots

Bumps conformance to 0.0.9

Sigstore cli now uses base64 instead of raw bytes when writing and reading signature files


vlsi (approved):
nice

PR #602, opened 2024-01-08
gpg and sigstore sign publications

loosebazooka:
fyi @vlsi I removed the stage/vote plugin because it seemed a bit heavy weight for what I was looking for. Just simple signing.

In my tests, also these setup signing .sigstore with PGP. I'm not sure what I'm misisng.

### DIFF
--- a/build-logic/publishing/build.gradle.kts
+++ b/build-logic/publishing/build.gradle.kts
@@ -10,5 +10,6 @@ dependencies {
     implementation(project(":basics"))
     implementation(project(":jvm"))
     implementation("dev.sigstore.build-logic:gradle-plugin")
+    implementation("dev.sigstore:sigstore-gradle-sign-plugin:0.5.0")
     implementation("com.gradle.plugin-publish:com.gradle.plugin-publish.gradle.plugin:1.2.1")
 }
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -1,35 +1,11 @@
 plugins {
     id("build-logic.root-build")
-    // It does not support participating in precompiled script plugins
-    id("com.github.vlsi.stage-vote-release") version "1.90"
     // The Kotlin Gradle plugin was loaded multiple times in different subprojects, which is not supported and may break the build.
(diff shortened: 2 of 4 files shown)


build-logic/publishing/build.gradle.kts
```diff
@@ -10,5 +10,6 @@ dependencies {
     implementation(project(":basics"))
     implementation(project(":jvm"))
     implementation("dev.sigstore.build-logic:gradle-plugin")
+    implementation("dev.sigstore:sigstore-gradle-sign-plugin:0.5.0")
```
vlsi, line 13:
```suggestion
    implementation("dev.sigstore.sign:dev.sigstore.sign.gradle.plugin:0.5.0")
```

This way the artifact name and group matches the plugin id

build.gradle.kts
```diff
@@ -1,35 +1,5 @@
...
-
-        Git SHA: ${it.gitSha}
-        Staging repository: ${it.nexusRepositoryUri}
-        """.trimIndent()
-    }
-}
-
-allprojects {
-    version = rootProject.version
-}
```
vlsi, line 35: This was conditionally adding `-SNAPSHOT` when building non-release version. I believe we would like to keep building `-SNAPSHOT` versions by default.

loosebazooka: ah gotcha. I can fix that.

vlsi:
> Another workaround could be removing the unwanted signatures by calling afterEvaluate { Sign.signatures.remove(...) }

We might factor and ship this as a helper plugin so the users could produce .sigstore without .asc as well.

loosebazooka:
unfortunately the artifacts don't all to seem be available in the afterEvaluate block. I can only modify the .module and .pom files?

loosebazooka:
I'm also not sure how to convert PublicationArtifact to PublishArtifact

loosebazooka:
@vlsi, how about we merge this for now and work on the ignoring .sigstore files in an update?

vlsi (approved):

PR #604, opened 2024-01-10
Custom signing task to ignore sigstore sigs

loosebazooka:
This is the best I could do for this :shrug: ?

@ljacomet @vlsi wdyt?

It's basically re-implementing parts of the sign extension/task internals to ignore sigstore files. It not fully featured (doesn't handle whenObjectRemoved, etc)

### DIFF
--- a/build-logic/publishing/src/main/kotlin/build-logic.java-published-library.gradle.kts
+++ b/build-logic/publishing/src/main/kotlin/build-logic.java-published-library.gradle.kts
@@ -1,3 +1,8 @@
+import org.codehaus.groovy.runtime.StringGroovyMethods
+import org.gradle.api.publish.internal.PublicationInternal
+import org.gradle.api.publish.internal.PublicationInternal.DerivedArtifact
+import org.gradle.api.publish.maven.internal.publication.MavenPublicationInternal
+
 plugins {
     id("build-logic.repositories")
     id("build-logic.java-library")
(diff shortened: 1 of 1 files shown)


vlsi:
I would rather expose it as `dev.sigstore.unsign-sigstore-pgp` plugin which removes pgp signatures if added by Gradle.
Then everybody can use the plugin as they see fit as long as https://github.com/gradle/gradle/issues/26760 is still open.

WDYT?

loosebazooka:
Sure, is it possible for it to just be a configuration setting on the sigstore sign plugin?

vlsi:
There might be use cases for not using sigstore's "sign everything with sigstore" `dev.sigstore.sign` plugin, however, users might still want to "unsign `.sigstore.asc`".

So the feature of "unsigning" is a well-defined one, and it fits for a new plugin just fine.

Of course, our umbrella `dev.sigstore.sign` plugin could automatically apply `dev.sigstore.unsign.pgp` (and allow an escape hatch), however, it makes perfect sense to factor `unsign` into its plugin.

See https://docs.gradle.org/current/userguide/designing_gradle_plugins.html#capabilities-vs-conventions for more details:

> One way to provide these quality criteria is to separate capabilities from conventions. In practice that means separating general-purpose functionality from pre-configured, opinionated functionality

"unsign pgp" is a capability while "dev.sigstore.sign" is a convention to "sign everything with sigstore and unsign sigstore.pgp"

loosebazooka:
gotcha, is this something you might have some bandwidth for right now?

PR #605, opened 2024-01-12
feat: remove .sigstore.asc signatures if added by signing plugin

vlsi:
#### Summary

By default, the plugin will remove .sigstore.asc.
Project property dev.sigstore.sign.remove.sigstore.asc=false would keep .sigstore.asc files if they are needed

Closes https://github.com/sigstore/sigstore-java/pull/604

#### Release Note

Remove `.sigstore.asc` from Gradle publications.

#### Documentation

NONE

### DIFF
--- a/sigstore-gradle/sigstore-gradle-sign-base-plugin/src/main/kotlin/dev/sigstore/sign/SigstoreSignExtension.kt
+++ b/sigstore-gradle/sigstore-gradle-sign-base-plugin/src/main/kotlin/dev/sigstore/sign/SigstoreSignExtension.kt
@@ -30,6 +30,8 @@ import org.gradle.kotlin.dsl.create
 import org.gradle.kotlin.dsl.named
 import org.gradle.kotlin.dsl.register
 import org.gradle.kotlin.dsl.the
+import org.gradle.kotlin.dsl.withType
+import org.gradle.plugins.signing.Sign
 import kotlin.collections.set
 
 abstract class SigstoreSignExtension(private val project: Project) {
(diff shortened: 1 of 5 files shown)


sigstore-gradle/sigstore-gradle-sign-base-plugin/src/main/kotlin/dev/sigstore/sign/SigstoreSignExtension.kt
```diff
@@ -92,6 +99,20 @@ abstract class SigstoreSignExtension(private val project: Project) {
                     publishableArtifact,
                     DefaultDerivedArtifactFile(project.tasks.named<DefaultTask>(signTask.name), signatureLocation)
                 ).apply { builtBy(signTask) }
+                // Gradle's signing plugin reacts on adding artifacts, and it might add .asc signature
+                // So we need to remove .sigstore.asc as it is unwanted in most of the cases
+                if (removeSigstoreAsc) {
+                    project.tasks.withType<Sign>()
+                        .matching { it.name.contains(publicationName, ignoreCase = true) }
```
loosebazooka, line 106: I don't know that this matters, but the task name generation is done with (https://github.com/gradle/gradle/blob/322c7dedcafea4c1302994f9ad012e9bb358bf83/platforms/software/signing/src/main/java/org/gradle/plugins/signing/SigningExtension.java#L415)

so maybe matching with that would be a tighter match?

Unless custom task names are being generated (but I'm not sure we could handle those anyway?)

loosebazooka (approved):

sigstore-gradle/sigstore-gradle-sign-base-plugin/src/main/kotlin/dev/sigstore/sign/SigstoreSignExtension.kt
vlsi: Do we need a tight match? This is more like a perf optimization to avoid excessive `collection.removeIf`.
I would rather keep code simpler and ask @ljacomet so Gradle adds something to configure exclusions from regular signing.

PR #607, opened 2024-01-19
Preserve raw signed json segments in TUF metadata

loosebazooka:
This should handle the issue of new fields like: https://github.com/sigstore/root-signing-staging/blob/fecfda76bb9a1721b37dd581f713b72e41270447/metadata/root_history/5.root.json#L23 being added within the signed block of tuf metadata.


patflynn (approved):
nice job!

PR #618, opened 2024-01-23
Release to maven central (sigstore-java only)

loosebazooka:
this still requires someone with access to dev.sigstore to complete the release on s01.oss.sonatype.org


.github/workflows/release-sigstore-java-from-tag.yaml
```diff
@@ -56,22 +48,30 @@ jobs:
...
-      - name: Build project
+      - name: Build, Sign and Release to Maven Central
         run: |
-          ./gradlew clean :sigstore-java:createReleaseBundle -Pversion=${{ github.event.inputs.release_version }} -Prelease -PskipSign
+          ./gradlew clean :sigstore-java:publishMavenJavaPublicationToSonatypeRepository -Prelease
+        env:
+          ORG_GRADLE_PROJECT_signingKey: ${{ secrets.PGP_PRIVATE_KEY }}
+          ORG_GRADLE_PROJECT_signingPassword: ${{ secrets.PGP_PASSPHRASE }}
+          ORG_GRADLE_PROJECT_sonatypeUsername: ${{ secrets.SONATYPE_USERNAME }}
+          ORG_GRADLE_PROJECT_sonatypePassword: ${{ secrets.SONATYPE_PASSWORD }}
```
szpak, line 58: A generic questions. I wonder, if it is "safer" to use system environment or `-P` in the Gradle calls?

System envs are populated across the system/container. On the other hand, `-P` params might be exposed on the process list.

build-logic/publishing/src/main/kotlin/build-logic.publish-to-central.gradle.kts
```diff
@@ -61,27 +59,11 @@ publishing {
...
-                from(publicationArtifact.file)
-            }
+    repositories {
+        maven {
+            name = "sonatype"
+            url = uri("https://s01.oss.sonatype.org/service/local/staging/deploy/maven2/")
+            credentials(PasswordCredentials::class)
         }
     }
 }
```
szpak, line 69: It's beyond the scope of that PR, but you might consider automatically publishing SNAPSHOT to Maven Central on every build from the default branch* (unless I miss and it is already being done).

* - some logic to build the version would be required.

.github/workflows/release-sigstore-java-from-tag.yaml
loosebazooka: Yeah, I mean that's probably true. I can use `-PsigningKey`, etc here instead

loosebazooka: oh wait, I don't think github actions allows you to access secrets directly.

szpak: You might be right. I'm not sure if `-PsigningKey=${{ secrets.PGP_PRIVATE_KEY }}` is supported. Even if, in the documentation, using envs is recommended, so probably their try to protect secrets at the "container" level (and keeping it as is might be a better idea).

loosebazooka:
@jku fyi

build-logic/publishing/src/main/kotlin/build-logic.publish-to-central.gradle.kts
loosebazooka: that's fine, I think we can do that in the future

patflynn (approved):

build-logic/publishing/src/main/kotlin/build-logic.publish-to-central.gradle.kts
szpak: Reported as https://github.com/sigstore/sigstore-java/issues/620

(but I can't "resolve" this thread)

PR #625, opened 2024-02-02
Be more permissive for signatures in TUF metadata

loosebazooka:
Don't fail if a signature doesn't validate, just move on to the next and as long as we meet the threshold we are fine


loosebazooka:
@jku fyi

jku (approved):
This looks correct to me

sigstore-java/src/main/java/dev/sigstore/tuf/Updater.java
```diff
@@ -221,13 +224,22 @@ void verifyDelegate(
...
             }
-          } catch (SignatureException e) {
-            throw new TufException(e);
+          } catch (SignatureException
+              | DecoderException
+              | NoSuchAlgorithmException
+              | InvalidKeyException e) {
+            log.fine(
+                "invalid signature '" + signature.getSignature() + "' for keyid '" + keyid + "'");
+            log.fine(e.getMessage());
```
vlsi, line 242:
```suggestion
            log.log(Level.FINE,
                "invalid signature ''{0}}'' for keyid ''{1}''", new Object[]{signature.getSignature(), keyid, e});
```

Note the repeated single quotes `'` so they convert to a single quote in log.

loosebazooka: how does the throwable work here? Is there some magic that adds it to the message? Or do we need to use the logger than accepts throwable explicitly?

loosebazooka: maybe more like `log.log(Level.FINE, "invalid signature: " + sig + " for keyid: " + keyid ", e)`

loosebazooka: nvm, I reworked this a bit.

sigstore-java/src/main/java/dev/sigstore/tuf/Updater.java
```diff
@@ -221,13 +225,26 @@ void verifyDelegate(
...
-          } catch (SignatureException e) {
-            throw new TufException(e);
+          } catch (SignatureException
+              | DecoderException
+              | NoSuchAlgorithmException
+              | InvalidKeyException e) {
+            log.log(
+                Level.FINE,
+                "TUF: ignored invalid signature: "
+                    + signature.getSignature()
```
vlsi, line 244: I am afraid it would be prone to log injection. The message will be passed through messageformat anyway. Wdyt?

vlsi: Yes

loosebazooka: alright cool lemme fix that

loosebazooka: Should we also just switch to using a slf4j or log4j?

vlsi: I would be -1..-0.5 re log4j as their backward compatibility is far from being solid (e.g. they completely ignore CVEs in log4j1, and they might be breaking backward compatibility betwen log4j2-api and log4j3-api)
It might be ok to use slf4j

loosebazooka: yeah, well I guess we can consider that. We don't do much logging anyway, so maybe jul is fine for now.

sigstore-java/src/main/java/dev/sigstore/tuf/Updater.java
```diff
@@ -221,13 +225,28 @@ void verifyDelegate(
...
               goodSigs.add(signature.getKeyId());
             }
-          } catch (SignatureException e) {
-            throw new TufException(e);
+          } catch (SignatureException
+              | DecoderException
+              | NoSuchAlgorithmException
+              | InvalidKeyException e) {
+            log.log(
+                Level.FINE,
```
patflynn, line 242: 'warn' maybe? seems like a data problem that you'd want to know about.

loosebazooka: I think these errors can occur naturally with no ill effect. Maybe we should separate expected from unexpected here? And log accordingly?

@jku presumably just the `""` case is expected?

jku: Without knowing details of the exceptions:
* In a generic TUF client a signature mismatch should not generate warnings visible to users by default
* since this is specifically a sigstore TUF client, you could use the knowledge that only empty strings are used as "signatures": I think I would not warn even in other signature mismatch cases though
* the other errors sound like issues that maybe should result in something visible to user

loosebazooka: alright moved things around.

sigstore-java/build.gradle.kts
```diff
@@ -57,6 +57,9 @@ dependencies {
     testImplementation("com.squareup.okhttp3:mockwebserver:4.12.0")
     testImplementation("net.sourceforge.htmlunit:htmlunit:2.70.0")
     testImplementation("org.eclipse.jetty:jetty-server:11.0.20")
+
+    testImplementation("io.github.netmikey.logunit:logunit-core:2.0.0")
+    testRuntimeOnly("io.github.netmikey.logunit:logunit-jul:2.0.0")
```
vlsi, line 62: Could you please add `io.github.netmikey.logunit` to `renovate.json` so we get synchronized version bumps by Renovate?

sigstore-java/src/test/java/dev/sigstore/tuf/UpdaterTest.java
```diff
@@ -69,6 +89,9 @@ class UpdaterTest {
   @TempDir Path localStorePath;
   @TempDir static Path localMirrorPath;
 
+  @RegisterExtension
+  LogCapturer logs = LogCapturer.create().captureForType(Updater.class, Level.DEBUG);
```
vlsi, line 93: nice

vlsi (changes requested):

bobcallaway (approved):

PR #643, opened 2024-02-22
Update protobuf-spec to 0.3.0

loosebazooka:


loosebazooka:
@vlsi if you have time, we could use a review on this.

vlsi (approved):

PR #646, opened 2024-02-27
Fix ed25519 key handling

loosebazooka:
Turns out, we need to handle that TUF allows ed25519 keys are raw 64 byte hex.

### DIFF
--- a/sigstore-java/src/main/java/dev/sigstore/encryption/Keys.java
+++ b/sigstore-java/src/main/java/dev/sigstore/encryption/Keys.java
@@ -34,6 +34,8 @@ import java.security.spec.X509EncodedKeySpec;
 import java.util.List;
 import org.bouncycastle.asn1.ASN1Integer;
 import org.bouncycastle.asn1.ASN1Sequence;
+import org.bouncycastle.asn1.edec.EdECObjectIdentifiers;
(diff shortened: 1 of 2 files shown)


sigstore-java/src/main/java/dev/sigstore/encryption/Keys.java
```diff
@@ -137,9 +139,24 @@ public static PublicKey constructTufPublicKey(byte[] contents, String scheme)
       case "ed25519":
         {
           final KeyFactory kf = KeyFactory.getInstance("Ed25519");
-          final X509EncodedKeySpec keySpec = new X509EncodedKeySpec(contents);
+          X509EncodedKeySpec keySpec;
+          // tuf allows raw keys only for ed25519 (non PEM):
+          // https://github.com/theupdateframework/specification/blob/master/tuf-spec.md#key-objects--file-formats-keys
```
vlsi, line 144: I would prefer permalinks rather than links to master branch which might remove or rename file over time

vlsi (approved):

sigstore-java/src/main/java/dev/sigstore/encryption/Keys.java
loosebazooka: Oh right. Yeah I'll fix that

vlsi (approved):

PR #656, opened 2024-03-11
TUF allows optional hashes and length on snapshots

loosebazooka:
This is defined in the spec here:
https://github.com/theupdateframework/specification/blob/53bb11dcd448cc9bd7c657eb3bbd72b073c683e6/tuf-spec.md#file-formats-snapshotjson--file-formats-snapshot


sigstore-java/src/main/java/dev/sigstore/tuf/model/SnapshotMeta.java
```diff
@@ -29,6 +31,8 @@
 @Value.Immutable
 public interface SnapshotMeta extends TufMeta {
 
+  Integer DEFAULT_MAX_LENGTH = 2000000;
```
patflynn, line 34: just curious where the default comes from?

patflynn:
don't we need some test data without the values?

sigstore-java/src/main/java/dev/sigstore/tuf/model/SnapshotMeta.java
loosebazooka: oh guess I never added the comment, it comes from the python-tuf defaults.

loosebazooka:
> don't we need some test data without the values?

lol yeah, we definitely do. Been looking through how you generated the old tuf data

loosebazooka:
kk tests update. 

added @jku for sanity check

jku (approved):
Looks fine to me.

patflynn (approved):

PR #657, opened 2024-03-13
Handle parsing checkpoints from a rekor entry

loosebazooka:
This does not do validation on the checkpoint yet

That should be added in a followup -- this PR is already kinda big


sigstore-java/src/main/java/dev/sigstore/rekor/client/Checkpoints.java
```diff
@@ -0,0 +1,98 @@
...
+    return ImmutableCheckpoint.builder()
+        .origin(origin)
+        .size(size)
+        .base64Hash(base64Hash)
+        .addAllSignatures(signatures)
+        .build();
+  }
+
+  static CheckpointSignature sigFrom(String signatureLine) throws RekorParseException {
+    var sigBlock = Pattern.compile("\\u2014 (\\S+) (\\S+)");
```
vlsi, line 79: WDYT of moving it to `private static final` field?

sigstore-java/src/main/java/dev/sigstore/rekor/client/Checkpoints.java
```diff
@@ -0,0 +1,98 @@
...
+    if (headers.size() < 3) {
+      throw new RekorParseException("Checkpoint header must contain at least 3 lines");
+    }
+
+    var origin = headers.get(0);
+    long size;
+    try {
+      size = Long.parseLong(headers.get(1));
+    } catch (NumberFormatException nfe) {
+      throw new RekorParseException("Checkpoint header attribute size must be a number");
```
vlsi, line 53: Should the error message include the problematic value?

sigstore-java/src/main/java/dev/sigstore/rekor/client/Checkpoints.java
```diff
@@ -0,0 +1,98 @@
...
+      throw new RekorParseException("Checkpoint body must contain at least one signature");
+    }
+    if (!data.endsWith("\n")) {
+      throw new RekorParseException("Checkpoint signature section must end with newline");
+    }
+
+    List<CheckpointSignature> signatures = new ArrayList<>();
+    for (String sig : data.lines().collect(Collectors.toList())) {
+      signatures.add(sigFrom(sig));
+    }
```
vlsi, line 68: This might work as well (it is `for (... collect(toList())` that looks suspicious):

```suggestion
    List<CheckpointSignature> signatures = data.lines().map(::sigFrom).collect(Collectors.toList())
```

sigstore-java/src/main/java/dev/sigstore/rekor/client/Checkpoints.java
```diff
@@ -0,0 +1,98 @@
...
+        .addAllSignatures(signatures)
+        .build();
+  }
+
+  static CheckpointSignature sigFrom(String signatureLine) throws RekorParseException {
+    var sigBlock = Pattern.compile("\\u2014 (\\S+) (\\S+)");
+    var m = sigBlock.matcher(signatureLine);
+    if (!m.find()) {
+      // really shouldn't be getting here because we check
+      throw new RekorParseException("Checkpoint signature was invalid or improperly formatted");
```
vlsi, line 83: It would probably make sense to add a more specific error condition. For instance: "Checkpoint signature should start with ..."

I know it is unlikely to get all these failures, however, better messages would make debugging easier.

sigstore-java/src/main/java/dev/sigstore/rekor/client/Checkpoints.java
```diff
@@ -0,0 +1,98 @@
...
+    }
+    var identity = m.group(1);
+    var keySig = Base64.getDecoder().decode(m.group(2));
+    if (keySig.length < 5) {
+      throw new RekorParseException("Checkpoint signature contains too few bytes");
```
vlsi, line 88: `expected 5 bytes, got ...`?

vlsi (changes requested):

sigstore-java/src/main/java/dev/sigstore/rekor/client/Checkpoints.java
loosebazooka: Yeah the problem is the checked exception in the `sigFrom` call.

sigstore-java/src/main/java/dev/sigstore/rekor/client/Checkpoints.java
loosebazooka: Oh also this comment (line 82) is outdated, I removed that older check.

sigstore-java/src/main/java/dev/sigstore/rekor/client/Checkpoints.java
loosebazooka: I think it's more expected more than 5 bytes (4bytes of key hint + Xbytes of signature)

PR #671, opened 2024-03-27
Rework signing initialization

loosebazooka:
Make the signing clients configured with URIs instead of objects from the trusted_root.json. This code still uses the URIs from the trusted_root.json to populate those values, but this is a setup towards using the SigningConfiguration type that is coming in the future.

### DIFF
--- a/sigstore-java/src/main/java/dev/sigstore/KeylessSigner.java
+++ b/sigstore-java/src/main/java/dev/sigstore/KeylessSigner.java
@@ -191,9 +191,11 @@ public class KeylessSigner implements AutoCloseable {
       Preconditions.checkNotNull(sigstoreTufClient, "sigstoreTufClient");
       sigstoreTufClient.update();
       var trustedRoot = sigstoreTufClient.getSigstoreTrustedRoot();
-      var fulcioClient = FulcioClient.builder().setCertificateAuthority(trustedRoot).build();
+      var fulcioClient =
+          FulcioClient.builder().setUri(trustedRoot.getCAs().current().getUri()).build();
       var fulcioVerifier = FulcioVerifier.newFulcioVerifier(trustedRoot);
-      var rekorClient = RekorClient.builder().setTransparencyLog(trustedRoot).build();
+      var rekorClient =
+          RekorClient.builder().setUri(trustedRoot.getTLogs().current().getBaseUrl()).build();
       var rekorVerifier = RekorVerifier.newRekorVerifier(trustedRoot);
       return new KeylessSigner(
           fulcioClient,
@@ -354,7 +356,7 @@ public class KeylessSigner implements AutoCloseable {
         }
       }
 
-      CertPath signingCert =
+      CertPath renewedSigningCert =
           fulcioClient.signingCertificate(
               CertificateRequest.newCertificateRequest(
                   signer.getPublicKey(),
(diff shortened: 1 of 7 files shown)


sigstore-java/src/main/java/dev/sigstore/KeylessSigner.java
```diff
@@ -191,9 +191,11 @@ public KeylessSigner build()
       Preconditions.checkNotNull(sigstoreTufClient, "sigstoreTufClient");
       sigstoreTufClient.update();
       var trustedRoot = sigstoreTufClient.getSigstoreTrustedRoot();
-      var fulcioClient = FulcioClient.builder().setCertificateAuthority(trustedRoot).build();
+      var fulcioClient =
+          FulcioClient.builder().setUri(trustedRoot.getCAs().current().getUri()).build();
       var fulcioVerifier = FulcioVerifier.newFulcioVerifier(trustedRoot);
-      var rekorClient = RekorClient.builder().setTransparencyLog(trustedRoot).build();
+      var rekorClient =
+          RekorClient.builder().setUri(trustedRoot.getTLogs().current().getBaseUrl()).build();
```
loosebazooka, line 198: this bit will be modified in the future to be initalized from the signing init config.

haydentherapper: How does this handle if multiple logs or CAs are active? Does `current` take the first?

sigstore-java/src/main/java/dev/sigstore/KeylessVerifier.java
```diff
@@ -221,15 +231,17 @@ private RekorEntry getEntryFromRekor(
     }
     Optional<RekorEntry> rekorEntry;
 
-    // attempt to grab the rekord from the rekor instance
+    // attempt to grab the rekord from all known rekor instances
     try {
-      rekorEntry = rekorClient.getEntry(hashedRekordRequest);
-      if (rekorEntry.isEmpty()) {
-        throw new KeylessVerificationException("Rekor entry was not found");
+      for (var rekorClient : rekorClients) {
```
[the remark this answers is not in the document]
haydentherapper: If you're doing an online lookup currently, this is fine, but +1 on moving towards requiring offline verification. I do agree you might want to limit the lookups or make it configurable.

sigstore-java/src/main/java/dev/sigstore/KeylessVerifier.java
```diff
@@ -72,9 +77,14 @@ public KeylessVerifier build()
...
       var fulcioVerifier = FulcioVerifier.newFulcioVerifier(trustedRoot);
-      var rekorClient = RekorClient.builder().setTransparencyLog(trustedRoot).build();
       var rekorVerifier = RekorVerifier.newRekorVerifier(trustedRoot);
-      return new KeylessVerifier(fulcioVerifier, rekorClient, rekorVerifier);
+      var rekorClients =
+          trustedRoot.getTLogs().getTransparencyLogs().stream()
+              .map(TransparencyLog::getBaseUrl)
+              .distinct()
+              .map(uri -> RekorClient.builder().setUri(uri).build())
+              .collect(Collectors.toList());
```
haydentherapper, line 86: Does this preserve order so the newer logs are first? This might help when it comes to searching all logs (unless you're mostly verifying old artifacts).

sigstore-java/src/main/java/dev/sigstore/fulcio/client/FulcioClient.java
```diff
@@ -45,19 +43,19 @@
...
-  private FulcioClient(HttpParams httpParams, CertificateAuthority certificateAuthority) {
-    this.certificateAuthority = certificateAuthority;
+  private FulcioClient(HttpParams httpParams, URI uri) {
+    this.uri = uri;
     this.httpParams = httpParams;
   }
 
   public static class Builder {
-    private CertificateAuthority certificateAuthority;
+    private URI uri = URI.create("https://fulcio.sigstore.dev");
```
haydentherapper, line 58: How will staging be handled? Or is this just a default?

sigstore-java/src/main/java/dev/sigstore/KeylessSigner.java
loosebazooka: Yeah that's just an artifact of the older code that was initializing under the assumption there was only one CA. Once https://github.com/sigstore/protobuf-specs/pull/277 messages are created, we'll init from that.

sigstore-java/src/main/java/dev/sigstore/fulcio/client/FulcioClient.java
loosebazooka: This is just a default. The builder allows this to be configurable.

haydentherapper (approved):

haydentherapper (approved):

PR #674, opened 2024-04-05
Generate java client for proto at build

loosebazooka:
Stop using pre-generated protobuf client libraries, just use the proto files directly when generating clients. This allows us to update proto-gen versions without worrying about conflicts (We were geneating client for fulcio.proto locally while using pre generated proto clients for protobuf-spec)

A followup to this will be to remove the generator from https://github.com/sigstore/protobuf-specs and only publish the .proto files in a jar.

Also supersedes #669

### DIFF
--- a/sigstore-java/build.gradle.kts
+++ b/sigstore-java/build.gradle.kts
@@ -20,13 +20,11 @@ dependencies {
 
     implementation("io.github.erdtman:java-json-canonicalization:1.1")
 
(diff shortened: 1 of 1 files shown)


haydentherapper (approved):

loosebazooka:
I think fuzzing instrumentation is broken on protobuf 4, we'll just have to leave it broken while we figure it out. This PR should still go in though.

haydentherapper (approved):

PR #683, opened 2024-04-26
make signature extension .sigstore.json

loosebazooka:
This was the best way I could do this (@vlsi perhaps not kotlin style). There seems to be a way that gradle is generating extensions for derived artifacts that isn't compatible with our double `.` extension: https://github.com/gradle/gradle/issues/28969 but I've worked around that for now.

### DIFF
--- a/sigstore-gradle/sigstore-gradle-sign-base-plugin/src/main/kotlin/dev/sigstore/sign/SigstoreSignExtension.kt
+++ b/sigstore-gradle/sigstore-gradle-sign-base-plugin/src/main/kotlin/dev/sigstore/sign/SigstoreSignExtension.kt
@@ -26,6 +26,9 @@ import org.gradle.api.provider.Property
 import org.gradle.api.publish.Publication
 import org.gradle.api.publish.PublicationArtifact
 import org.gradle.api.publish.internal.PublicationInternal
+import org.gradle.api.publish.maven.MavenArtifact
+import org.gradle.api.publish.maven.internal.artifact.AbstractMavenArtifact
+import org.gradle.api.publish.maven.internal.artifact.DerivedMavenArtifact
 import org.gradle.kotlin.dsl.create
 import org.gradle.kotlin.dsl.named
 import org.gradle.kotlin.dsl.register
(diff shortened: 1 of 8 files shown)


sigstore-gradle/sigstore-gradle-sign-base-plugin/src/main/kotlin/dev/sigstore/sign/SigstoreSignExtension.kt
```diff
@@ -88,28 +91,34 @@ abstract class SigstoreSignExtension(private val project: Project) {
         val artifacts = mutableMapOf<PublicationArtifact, T>()
         publication.allPublishableArtifacts {
             val publishableArtifact = this
-            if (file.extension !in listOf("asc", SigstoreSignature.EXTENSION)) {
+            if (!file.name.endsWith(".asc") and !file.name.endsWith(SigstoreSignature.EXTENSION)) {
```
vlsi, line 94:
```suggestion
            if (!file.name.endsWith(".asc") && !file.name.endsWith(SigstoreSignature.EXTENSION)) {
```

`bool and bool` doesn't short-curcuit, so regular `&&` is the way to go

sigstore-gradle/sigstore-gradle-sign-base-plugin/src/main/kotlin/dev/sigstore/sign/SigstoreSignature.kt
```diff
@@ -25,7 +25,7 @@ import javax.inject.Inject
 
 abstract class SigstoreSignature @Inject constructor(private val name: String) : Named {
     companion object {
-        const val EXTENSION = "sigstore"
+        const val EXTENSION = ".sigstore.json"
```
vlsi, line 28: Apache commons-io does not include dot in the extension: https://commons.apache.org/proper/commons-io/javadocs/api-2.5/org/apache/commons/io/FilenameUtils.html#getExtension(java.lang.String)

Maven does not include the dot in the extension too: https://maven.apache.org/repositories/artifacts.html

`EXTENSION` is public, so we might exclude the dot from it. WDYT?

loosebazooka: Sounds good. I'll adjust.

sigstore-gradle/sigstore-gradle-sign-base-plugin/src/main/kotlin/dev/sigstore/sign/SigstoreSignExtension.kt
loosebazooka: Sgtm

loosebazooka:
@vlsi when you get a chance, can we do an another look over and then we can maybe start another release

vlsi (approved):

PR #684, opened 2024-04-29
add staging to conformance

loosebazooka:
staging has been a bit flaky lately, so make it a separate job so it doesn't block merges

### DIFF
--- a/.github/workflows/conformance.yml
+++ b/.github/workflows/conformance.yml
@@ -5,12 +5,15 @@ on:
     branches:
       - '**'
   pull_request:
+  # TODO: add cron
 
 jobs:
   conformance:
     strategy:
+      max-parallel: 1
       matrix:
         java-version: [11, 17]
+        sigstore-env: [production, staging]
       fail-fast: false
 
     runs-on: ubuntu-latest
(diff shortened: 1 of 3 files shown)


.github/workflows/conformance.yml
```diff
@@ -5,12 +5,14 @@ on:
...
       - '**'
   pull_request:
+  # TODO: add cron
 
 jobs:
   conformance:
     strategy:
       matrix:
         java-version: [11, 17]
+        sigstore-env: [production, staging]
```
jku, line 15: At least in context of current latency issues you may want to avoid two staging jobs concurrently... Maybe worth considering if the two java versions give you value here considering ci.yaml already tests two versions?

jku (approved):
:+1:

.github/workflows/conformance.yml
loosebazooka: I can make them non concurrent

loosebazooka: there are security library differences between 11 and 17 that I want to potentially fail if it comes to it.

loosebazooka: These finish pretty quick so just made them run non-parallel

jku (approved):

PR #688, opened 2024-05-02
Use new oci compat mediatype for bundles

loosebazooka:


sigstore-java/src/main/java/dev/sigstore/bundle/BundleFactoryInternal.java
```diff
@@ -59,8 +59,10 @@ class BundleFactoryInternal {
   private static final String BUNDLE_V_0_1 = "application/vnd.dev.sigstore.bundle+json;version=0.1";
   private static final String BUNDLE_V_0_2 = "application/vnd.dev.sigstore.bundle+json;version=0.2";
   private static final String BUNDLE_V_0_3 = "application/vnd.dev.sigstore.bundle+json;version=0.3";
+  // media_type format switch: https://github.com/sigstore/protobuf-specs/pull/279
+  private static final String BUNDLE_V_0_3_1 = "application/vnd.dev.sigstore.bundle.v0.3+json";
```
haydentherapper, line 63: To confirm, is there anywhere in the verification path where you need to treat `BUNDLE_V_0_3_1` and `BUNDLE_V_0_3` as equivalent bundles?

haydentherapper (approved):

sigstore-java/src/main/java/dev/sigstore/bundle/BundleFactoryInternal.java
loosebazooka: equivalent? We just parse them or we dont.

loosebazooka: The reality of this is we only have a single parser based on the protobuf spec build at v0.3.1, we're pretty permissive with the bundle mediatype. We don't punish v2, v1 bundles that use v3 features. If the v3 parser can parse it, we're okay and we use it.

loosebazooka: In fact, we ignore that v1 allows no inclusion proof. Those bundles are invalid as far as we're concerned.

haydentherapper (approved):

PR #709, opened 2024-05-22
Remove most compat apis

loosebazooka:
KeylessSignature is gone, just use Bundle now. Most APIs are changed to treat Bundle as the primary signing material object.

`signFile2` remains, but should be removed on the next release


patflynn (dismissed):

loosebazooka:
hrmm, I did something wrong

patflynn (approved):

PR #723, opened 2024-05-28
Allow cli to run with alt tuf remote repository

loosebazooka:
@jku presumably this should allow better conformance for the tuf repo. It does have separate flags for prod and staging (which allows the cli to be more robust), but it's not impossible to just use a single flag.


sigstore-cli/src/main/java/dev/sigstore/cli/Sign.java
```diff
@@ -40,12 +44,29 @@ public class Sign implements Callable<Integer> {
...
+        description = "use public good with a tuf remote repository override",
+        required = false)
+    String publicGoodWithTufUrlOverride;
+
+    @Option(
+        names = {"--staging-with-tuf-url-override"},
+        description = "use staging with a tuf remote repository override",
+        required = false)
+    String stagingWithTufUrlOverride;
+  }
```
loosebazooka, line 69: @jku there are the flags that allow changing the tuf url

jku (approved):
:+1: if you're ok putting this in the conformance cli, it does look like it allows me to easily include sigstore-java in the root-signing test workflow...  

(I'm not promising that root-signing tests will include "all" clients forever, but it would be nice to have for the near to mid-term...)

loosebazooka:
yeah the cli is whatever we want it to be. Just an interface into core functionality. We never release it.

vlsi (approved):

PR #729, opened 2024-06-04
Verification options

loosebazooka:
This makes it easier to verify things like github idtokens that change the SAN on each release tag.

~~needs #730 and #731~~

Turns out there was OID processing that changed because old fulcio OIDs were raw strings and new ones are DER Encoded UTF8. see: https://github.com/sigstore/fulcio/blob/main/docs/oid-info.md#extension-values

PR #730, opened 2024-06-05
Add StringMatchable interface

loosebazooka:
part of #729 

Makes it easier to either direct match strings or regex when comparing certificate OIDs (or really any string -- but this is what we use it for)

### DIFF
--- a/sigstore-java/src/main/java/dev/sigstore/strings/StringMatcher.java
+++ b/sigstore-java/src/main/java/dev/sigstore/strings/StringMatcher.java
@@ -1,0 +1,78 @@
+/*
+ * Copyright 2024 The Sigstore Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package dev.sigstore.strings;
+
+import java.util.Objects;
+import java.util.function.Predicate;
+import java.util.regex.Pattern;
+import java.util.regex.PatternSyntaxException;
+
+/**
+ * An interface for allowing direct string matching or regular expressions. Use the static factory
+ * {@link #string(String)} or {@link #regex(String)} to instantiate the appropriate matcher. Custom
+ * implementations should override {@link Object#toString} for better error reporting.
+ */
+public interface StringMatcher extends Predicate<String> {
+
+  /** Create a matcher for string equality. */
+  static StringMatcher string(String string) {
+    Objects.requireNonNull(string, "string matcher cannot be initialized with null string");
+    return new StringMatcher() {
+      @Override
+      public boolean test(String target) {
+        return string.equals(target);
+      }
+
+      @Override
+      public String toString() {
+        return "'String: " + string + "'";
+      }
+    };
+  }
+
+  /**
+   * Create a matcher using regular expressions. Regex matching ignores null values and returns
+   * false instead of erroring.
+   *
+   * @param string the input pattern
+   * @return a regex based instance
+   * @throws RegexSyntaxException if the input pattern is not valid regex. This is a runtime
+   *     exception and probably should be handled
+   */
+  static StringMatcher regex(String string) throws RegexSyntaxException {
+    Objects.requireNonNull(string, "string matcher cannot be initialized with null regex");
+    Pattern pattern;
+    try {
+      pattern = Pattern.compile(string);
+    } catch (PatternSyntaxException ex) {
+      throw new RegexSyntaxException("Could not parse regex: '" + string + "'", ex);
+    }
+    return new StringMatcher() {
+      @Override
+      public boolean test(String target) {
+        if (target == null) {
+          return false;
+        }
+        return pattern.matcher(target).matches();
+      }
+
+      @Override
+      public String toString() {
+        return "'RegEx: " + pattern + "'";
+      }
+    };
+  }
+}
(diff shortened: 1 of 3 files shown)


sigstore-java/src/main/java/dev/sigstore/StringMatchable.java
```diff
@@ -0,0 +1,66 @@
...
+package dev.sigstore;
+
+import com.google.common.base.Preconditions;
+import java.util.regex.Pattern;
+
+/**
+ * An interface for allowing direct string matching or regular expressions. Use the static factory
+ * {@link #of(String)} or {@link #of(Pattern)} to instantiate the appropriate matcher.
+ */
+public interface StringMatchable {
```
vlsi, line 25: The naming sounds a bit odd. Guava uses `CharMatcher implements Predicate<Character>`

sigstore-java/src/main/java/dev/sigstore/StringMatchable.java
```diff
@@ -0,0 +1,66 @@
...
+
+  boolean matches(String target);
+
+  /** The input provided to instantiate the string matchable. */
+  String source();
```
vlsi, line 30: What is the purpose of `source()` if the user can't tell if the resulting string is a regexp or literal?

loosebazooka: That's a good point, I can clarify that.

loosebazooka:
> My curiosity is peaked. I'm struggling to see what weight this abstraction is carrying over using .equals or .pattern directly (or just wrapping them with a static method).

I wanted to be able to specify if any comparisons that were happening in the library were regexes or plain strings. For example 

My policy could be accept any signatures from identity: `https://github.com/org/project/.github/workflows/release.yaml@refs/tags/v1.4.0`, but essentially that only applies to the release `1.4.0`
Instead my policy could accept signatures from identity: `https://github\\.com/org/project/\\.github/workflows/release\\.yaml@refs/tags/v\\d+\\.\\d+\\.\\d+` which now doesn't need constant modification.

There's not a great way to know deep in the library if a regex comparison or plain string comparison needs to happen, so just using equals or pattern directly isn't possible.

So it was an option between providing a convention for regexes or just make them separate types (mimic'ing a oneOf). I went with the interface abstraction so that it matches what the other libraries are doing -- make them explicit and separate.

sigstore-java/src/main/java/dev/sigstore/StringMatchable.java
loosebazooka: I had it as a functional interface at first, but it was hard to provide good error messaging when things didn't match. But the impl of source can make that more clear.

sigstore-java/src/main/java/dev/sigstore/StringMatchable.java
loosebazooka: I mean this matches more the interface naming (vs class naming) `Runnable`, `Callable`, etc. Vs `CharMatcher` which is an abstract class? I'm not super tied to this name, I just want this functionality.

sigstore-java/src/main/java/dev/sigstore/StringMatchable.java
vlsi: `source()` can be confusing if the string contains special characters.
For instance, consider `.of(".*")`.
The source would look like `.*`, and everybody would think it is a regexp.

I would rather have `.toString()` so the value represents better in the debugger.

sigstore-java/src/main/java/dev/sigstore/StringMatchable.java
vlsi: Compare with `java.util.function.Predicate` (vs `Predicatable` or `Testable`).

Frankly, this sounds like `StringPredicate extends Predicate<String>`.
At the same time, if protobuf-specs include a data-class-like (`regexp` and `literal` values), then we should probably expose it in the same way here.
Of course, `sealed interface` is not yet available, however we can expose it with protobuf-generated classes or just two case classes like `LiteralId` and `RegexId`.

loosebazooka: I don't think protobuf generated classes will be in the API. So we don't want to do that, I can rename a lot of these maybe. Lemme play around with it.

loosebazooka:
okay round 2, cleaned this up a lot. @patflynn @vlsi

sigstore-java/src/main/java/dev/sigstore/strings/StringMatcher.java
```diff
@@ -0,0 +1,77 @@
...
+      pattern = Pattern.compile(string);
+    } catch (PatternSyntaxException ex) {
+      throw new RegexSyntaxException("Could not parse regex: '" + string + "'", ex);
+    }
+    return new StringMatcher() {
+      @Override
+      public boolean test(String target) {
+        if (target == null) {
+          return false;
+        }
```
vlsi, line 67: Is this "null is silently ignored" behaviour documented?

vlsi (approved):

sigstore-java/src/main/java/dev/sigstore/strings/StringMatcher.java
loosebazooka: will document.

loosebazooka:
sorry @vlsi need a rereview unfortunately

vlsi (approved):

loosebazooka:
Once again, I am an idiot. My code format changes weren't committed. Sorry

sigstore-java/src/main/java/dev/sigstore/strings/StringMatcher.java
```diff
@@ -0,0 +1,78 @@
...
+   * Create a matcher using regular expressions. Regex matching ignores null values and returns
+   * false instead of erroring.
+   *
+   * @param string the input pattern
+   * @return a regex based instance
+   * @throws RegexSyntaxException if the input pattern is not valid regex. This is a runtime
+   *     exception and probably should be handled
+   */
+  static StringMatcher regex(String string) throws RegexSyntaxException {
+    Preconditions.checkNotNull(string);
```
vlsi, line 56: Technically, if you remove `Preconditions.checkNotNull(string);`, then Java 14+ would show helpful NPEs which are likely better than a generic failure from Preconditions

vlsi: Frankly, `Objects.requireNonNull(string, "regex pattern")` would indeed add value to the exception message.
However, I think `Objects.requireNonNull(string)` and `Preconditions.checkNotNull(string);` do not add much, and they should probably not be used

vlsi (approved):

PR #731, opened 2024-06-05
Use Matchers in OidcIdentity matching

loosebazooka:
Allows checking if signers match a string directly or a regex. Useful to limit which signers are used by config.


sigstore-java/src/test/java/dev/sigstore/KeylessSignerTest.java
```diff
@@ -127,20 +127,19 @@ public void sign_failGithubOidcCheck() throws Exception {
...
                     OidcIdentity.of(
-                        expectedGithubSubject, "https://token.actions.githubusercontent.com"),
-                    OidcIdentity.of("some@other.com", "https://accounts.other.com")))
+                        StringMatchable.of(
+                            Pattern.compile(
+                                "https://github\\.com/sigstore/sigstore-java/\\.github/workflows/.*\\.yaml@.*")),
+                        StringMatchable.of("https://token.actions.githubusercontent.com")),
+                    OidcIdentity.of(
+                        StringMatchable.of("some@other.com"),
+                        StringMatchable.of("https://accounts.other.com"))))
```
vlsi, line 142: Have you considered `allowedOidcIdentities(List<? extends Predicate<? super OidcIdentity>>)` instead?

sigstore-java/src/main/java/dev/sigstore/OidcIdentity.java
```diff
@@ -16,25 +16,25 @@
...
-        .build();
+  @Default
+  default boolean matches(OidcToken oidcToken) {
+    return getIdentity().matches(oidcToken.getSubjectAlternativeName())
+        && getIssuer().matches(oidcToken.getIssuer());
   }
 
   /** The user or machineId. */
-  String getIdentity();
+  StringMatchable getIdentity();
```
vlsi, line 36: Is it the same as `SubjectAlternativeName` which already seem to support regexp? https://github.com/sigstore/protobuf-specs/blob/3d01ad8fa60bdd0152bea2963b4d35282c2e263a/protos/sigstore_common.proto#L203-L210

loosebazooka: Yeah essentially, but doesn't actually implement anything. We still need a native Java datastructure to represent it.

vlsi: protobuf generates a Java representation, doesn't it? We do depend on protobuf anyway, so WDYT of using `SubjectAlternativeName` here?

loosebazooka: The java representation of protobuf always uses protobuf specific java classes, and I'd rather not expose those in our api.

vlsi: That is fair, however, we seem to expose Bundle, don't we?

loosebazooka: That's our own representation of bundle. Since the protobuf bundle uses all kinds of proto datastructures

loosebazooka:
@vlsi @patflynn this is be ready for review now.

patflynn (approved):

PR #734, opened 2024-06-06
chore: avoid generating protobuf(proto-google-common-protos) protobufs every time

vlsi:
`protobuf(...)` causes protoc to generate code for the listed proto files `compileProtoPath` adds the listed artifacts on the "include path", so we can reference third-party definitions and generate code for our proto files only.

Note: `com.google.protobuf` searches both `compileProtoPath` and implementation dependencies for .proto files.
`io.grpc:grpc-protobuf` already depends on `proto-google-common-protos`, so the build would work even without explicit dependency. However, we import `google/api` protos, so it would be nice to explicitly mention the used dependencies even though they are present as transitive.

### DIFF
--- a/sigstore-java/build.gradle.kts
+++ b/sigstore-java/build.gradle.kts
@@ -21,7 +21,9 @@ dependencies {
     implementation("io.github.erdtman:java-json-canonicalization:1.1")
 
     protobuf("dev.sigstore:protobuf-specs:0.3.2")
-    protobuf("com.google.api.grpc:proto-google-common-protos:2.39.1")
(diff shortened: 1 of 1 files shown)


vlsi:
I guess the reason for the failure is that `proto-google-common-protos` builds using an old `protoc` version.
We use `4.26.1` for `sigstore-java`, and `api-commons-protos` seem to use `3.14.0`: https://github.com/googleapis/api-common-protos/blob/3332dec527759859840a3a2ff108c67a54708130/Dockerfile#L7

loosebazooka:
gonna close this for now -- I was informed of this issue that might the be underlying problem: https://github.com/protocolbuffers/protobuf/issues/17247

PR #774, opened 2024-08-09
Add examples

loosebazooka:
needs #773


examples/hello-world/build.gradle.kts
```diff
@@ -0,0 +1,30 @@
...
+  mavenLocal() // for testing against dev builds
+  mavenCentral()
+}
+
+publishing {
+    publications {
+        create<MavenPublication>("maven") {
+            groupId = "com.example"
+            artifactId = "hello-world"
+            version = "1.0.0"
```
vlsi, line 19: Should this rather be at the top level?
I guess users should configure group, artifact, version anyway, so customizing the same at the publication level does not sound right.

loosebazooka: oh, yeah I guess this was kind of lazy on my part. Lemme update this.

examples/hello-world/gradle/libs.versions.toml
```diff
@@ -0,0 +1,2 @@
+# This file was generated by the Gradle 'init' task.
+# https://docs.gradle.org/current/userguide/platforms.html#sub::toml-dependencies-format
```
vlsi, line 2: WDYT of removing this file to make the example smaller?

patflynn (approved):

PR #811, opened 2024-09-29
Gradle plugin: Replace findProperty with Isolated Project compatible …

hfhbd:
…api gradleProperty

#### Summary
`findProperty` is not compatible with Gradle Isolated Projects, but `providers.gradleProperty` is. To test this change, you need an Isolated Projects compatible build, eg plain java library.

#### Release Note

Add support for Gradle Isolated Projects.

### DIFF
--- a/sigstore-gradle/sigstore-gradle-sign-base-plugin/src/main/kotlin/dev/sigstore/sign/SigstoreSignExtension.kt
+++ b/sigstore-gradle/sigstore-gradle-sign-base-plugin/src/main/kotlin/dev/sigstore/sign/SigstoreSignExtension.kt
@@ -84,7 +84,7 @@ abstract class SigstoreSignExtension(private val project: Project) {
         }
 
         val removeSigstoreAsc =
-            project.findProperty("dev.sigstore.sign.remove.sigstore.json.asc")?.toString()?.toBoolean() != false
+            project.providers.gradleProperty("dev.sigstore.sign.remove.sigstore.json.asc").map { it.toBoolean() }.orNull != false
 
         val publicationName = publication.name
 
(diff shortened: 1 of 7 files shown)


loosebazooka:
I guess we haven't had many forked PR created. Tests are failing on the availability of OIDC... which shouldn't be affected by this. I can override for the merge once we get approval.

vlsi:
Unfortunately, `providers.gradleProperty` is not the same as `findProperty` since `providers.gradleProperty` queries the values from the root project only which is conter-intuitive.
See https://github.com/gradle/gradle/issues/23572

Do you know if there's an alternative?
Do you know if [hasProperty](https://github.com/gradle/gradle/issues/23572#issuecomment-1487855035) supports isolated projects?

Could you add a test case so we know "isolated project" is supported or not?
Otherwise it could be we replace `findProperty` and gain nothing with it

hfhbd:
Yes, this is a change of the behavior but it is the only API compatible with isolated projects. hasProperty is also incompatible.

vlsi:
The mentioned property is an escape hatch, so it is probably fine to use `providers.gradleProperty`. Let's just merge it and be done with the PR.

A test with isolated projects would be very welcome.

sigstore-gradle/sigstore-gradle-sign-base-plugin/src/main/kotlin/dev/sigstore/sign/SigstoreSignExtension.kt
```diff
@@ -84,7 +84,7 @@ abstract class SigstoreSignExtension(private val project: Project) {
         }
 
         val removeSigstoreAsc =
-            project.findProperty("dev.sigstore.sign.remove.sigstore.json.asc")?.toString()?.toBoolean() != false
+            project.providers.gradleProperty("dev.sigstore.sign.remove.sigstore.json.asc").map { it.toBoolean() }.orNull != false
```
vlsi, line 87: WDYT of

```suggestion
            project.providers.gradleProperty("dev.sigstore.sign.remove.sigstore.json.asc").orNull?.toBoolean() != false
```

hfhbd: I don't care :D

hfhbd:
Any update?

loosebazooka:
Oh I didn't realize @vlsi approved.

loosebazooka:
can you rebase and update the pr?

hfhbd:
@loosebazooka rebased

vlsi (approved):

PR #825, opened 2024-10-04
adjustments to tuf client for conformance

loosebazooka:
This doesn't really bring us in line with how python-tuf works, but it does allow us to interface with the tuf-conformance suite a little better.

still todo:
- add cli for conformance
- maybe impl delegated targets


jku (approved):
Makes sense to me so far.

I would advice to not implement delegations on a whim: it's not trivial to get right and I would say it's much better to not support delegations than it is to support them partly.

PR #826, opened 2024-10-07
Move known roles in TUF

loosebazooka:
Roles were a bit too tightly coupled to the root roles, change this around a tiny bit so we can add delegations support

### DIFF
--- a/sigstore-java/src/main/java/dev/sigstore/tuf/MetaFetcher.java
+++ b/sigstore-java/src/main/java/dev/sigstore/tuf/MetaFetcher.java
@@ -15,9 +15,9 @@
  */
 package dev.sigstore.tuf;
 
-import dev.sigstore.tuf.model.Role;
 import dev.sigstore.tuf.model.Root;
 import dev.sigstore.tuf.model.SignedTufMeta;
+import dev.sigstore.tuf.model.TufMeta;
 import java.io.IOException;
 import java.util.Optional;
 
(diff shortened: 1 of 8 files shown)


sigstore-java/src/main/java/dev/sigstore/tuf/MetaFetcher.java
```diff
@@ -49,8 +49,8 @@ Optional<MetaFetchResult<Root>> getRootAtVersion(int version)
    * @throws FileExceedsMaxLengthException if the role meta at source exceeds client specified max
    *     size
    */
-  <T extends SignedTufMeta> Optional<MetaFetchResult<T>> getMeta(Role.Name name, Class<T> roleType)
-      throws IOException, FileExceedsMaxLengthException;
+  <T extends SignedTufMeta<? extends TufMeta>> Optional<MetaFetchResult<T>> getMeta(
+      String name, Class<T> roleType) throws IOException, FileExceedsMaxLengthException;
```
patflynn, line 53: It's probably obvious, but I don't quite see why we're relaxing from an enum to a string here.

vlsi: Is it because deserializing an unknown enum value would fail while we might want to skipping unknown roles instead?

loosebazooka: Cause you can grab a meta of a delegated target which names are not predefined?

patflynn: oh I see.  Should we rename the field to be more descriptive since it's just string?  I don't remember anymore what it represents. roleName? targetName?

vlsi: Just wondering: does it mean we must never use `enum` when deserializing input?
Should we replace `enum HashAlgorithm` with string as well?

loosebazooka: I don't think we were ever using the enum at parsing time. It looked like we were using the enum as a naming constraint within the code.

The description for Roles is here: https://theupdateframework.io/metadata/

But functionally, it's an object in an entity that is used to sign metadata.

So root contains roles for root, snapshot, timestamp and targets (the `RootRole` in this PR)
Target can contain roles for delegations (more targets) and these delegations can contain more roles for more delegations (creating some sort of tree).

patflynn (approved):

PR #827, opened 2024-10-10
Separate meta fetching from target fetching

loosebazooka:
Part of adjustments for conformance. Allows fetching meta and targets from different locations.

### DIFF
--- a/sigstore-java/src/main/java/dev/sigstore/tuf/Fetcher.java
+++ b/sigstore-java/src/main/java/dev/sigstore/tuf/Fetcher.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2023 The Sigstore Authors.
+ * Copyright 2024 The Sigstore Authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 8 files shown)


sigstore-java/src/main/java/dev/sigstore/tuf/Fetcher.java
```diff
@@ -0,0 +1,26 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package dev.sigstore.tuf;
+
+import java.io.IOException;
+
+public interface Fetcher {
```
patflynn, line 20: ResourceFetcher?

patflynn (approved):

sigstore-java/src/main/java/dev/sigstore/tuf/Fetcher.java
loosebazooka: should it be?

PR #836, opened 2024-10-25
Store meta state in memory

loosebazooka:
Delegated targets should be loaded at target search time, so keep state in memory so we can use it as necessary.

This PR doesn't does add delegated targets support, but this is part of the process.


sigstore-java/src/main/java/dev/sigstore/tuf/TrustedMeta.java
```diff
@@ -0,0 +1,114 @@
...
+  private TrustedMeta(MutableTufStore localStore) {
+    this.localStore = localStore;
+  }
+
+  static TrustedMeta newTrustedMeta(MutableTufStore localStore) {
+    return new TrustedMeta(localStore);
+  }
+
+  public void setRoot(Root root) throws IOException {
+    // call storeTrustedRoot instead of generic storeMeta because it does doesn't extra work
```
patflynn, line 43: ..does doesn't..

sigstore-java/src/main/java/dev/sigstore/tuf/TrustedMeta.java
```diff
@@ -0,0 +1,114 @@
...
+    localStore.storeTrustedRoot(root);
+    this.root = root;
+  }
+
+  public Root getRoot() throws IOException {
+    return findRoot().orElseThrow(() -> new IllegalStateException("No cached root to load"));
```
patflynn, line 49: maybe you can provide a little more context here as to why this would happen?

sigstore-java/src/main/java/dev/sigstore/tuf/TrustedMeta.java
```diff
@@ -0,0 +1,114 @@
...
+  }
+
+  public Optional<Root> findRoot() throws IOException {
+    if (root == null) {
+      root = localStore.loadTrustedRoot().orElse(null);
+    }
+    return Optional.ofNullable(root);
+  }
+
+  public void setTimestamp(Timestamp timestamp) throws IOException {
```
patflynn, line 59: not tempted to factor out the set/find/get logic?

sigstore-java/src/main/java/dev/sigstore/tuf/Updater.java
```diff
@@ -78,6 +83,7 @@ public class Updater {
     this.localStore = localStore;
     this.metaFetcher = metaFetcher;
     this.targetFetcher = targetFetcher;
+    this.trustedMeta = TrustedMeta.newTrustedMeta(localStore);
```
patflynn, line 86: since it's a read-through cache, could it be simpler to just invalidate it on update or something rather than having the updater have to know about it?

sigstore-java/src/main/java/dev/sigstore/tuf/TrustedMeta.java
loosebazooka: I was tempted but it meant gentrifying some of the getters and setters. I can do that

sigstore-java/src/main/java/dev/sigstore/tuf/Updater.java
loosebazooka: Oh like if the root updated?

loosebazooka: Yeah probably

sigstore-java/src/main/java/dev/sigstore/tuf/TrustedMeta.java
loosebazooka: I think you really only call "getRoot" if you know you should have a root cached somewhere. But I guess I can add context.

loosebazooka: cool

loosebazooka:
alright, reworked this whole thing

sigstore-java/src/main/java/dev/sigstore/tuf/MetaReader.java
```diff
@@ -0,0 +1,36 @@
...
+ * limitations under the License.
+ */
+package dev.sigstore.tuf;
+
+import dev.sigstore.tuf.model.SignedTufMeta;
+import dev.sigstore.tuf.model.TufMeta;
+import java.io.IOException;
+import java.util.Optional;
+
+public interface MetaReader {
```
patflynn, line 23: I'm trying to remember if the style guide no longer requires class level javadoc.

sigstore-java/src/main/java/dev/sigstore/tuf/MetaReader.java
```diff
@@ -0,0 +1,36 @@
...
+  /**
+   * Return a named metadata item if there is any.
+   *
+   * @param roleName the name of the role to load (root, timestamp, snapshot, targets, or a
+   *     delegated target role)
+   * @param tClass the class type
+   * @return an instance of the signed metadata for the role if it was found
+   * @throws IOException if an error occurs reading from the backing store
+   */
+  <T extends SignedTufMeta<? extends TufMeta>> Optional<T> findMeta(
```
patflynn, line 34: why name the interface *Reader and then use the verb findMeta instead of readMeta?

sigstore-java/src/main/java/dev/sigstore/tuf/MetaStore.java
```diff
@@ -15,28 +15,25 @@
...
-   *
-   * @param targetName the name of the target file to write (e.g. ctfe.pub)
-   * @param targetContents the content of the target file as bytes
-   * @throws IOException if an error occurs
-   */
-  void storeTargetFile(String targetName, byte[] targetContents) throws IOException;
+/** Interface that defined a mutable meta store functionality. */
+public interface MetaStore extends MetaReader {
+
+  String getIdentifier();
```
patflynn, line 26: javadoc

sigstore-java/src/main/java/dev/sigstore/tuf/MetaStore.java
```diff
@@ -15,28 +15,25 @@
...
 
   /**
-   * Generic method to store one of the {@link SignedTufMeta} resources in the local tuf store.
+   * Generic method to store one of the {@link SignedTufMeta} resources in the local tuf store. Do
+   * not use for Root role, use {@link #setRoot(Root)} instead.
```
patflynn, line 30: what happens if you do use it for Root?

loosebazooka: then you don't get N.root persisted (this is functionality that was already in there). I've just formalized it.

sigstore-java/src/main/java/dev/sigstore/tuf/MetaReader.java
loosebazooka: eh, added it in.

sigstore-java/src/main/java/dev/sigstore/tuf/PassthroughCacheMetaStore.java
```diff
@@ -0,0 +1,85 @@
...
+import java.io.IOException;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Optional;
+
+/** An in memory cache that will pass through to a provided local tuf store. */
+public class PassthroughCacheMetaStore implements MetaReader, MetaStore {
+  private final MetaStore localStore;
+  private final Map<String, SignedTufMeta<? extends TufMeta>> cache;
```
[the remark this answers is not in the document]
loosebazooka: Yeah it's a bit heavyweight, has eviction and what not. This is not a space limited cache.

patflynn (approved):

PR #837, opened 2024-10-25
only run errorpone in java 17+

loosebazooka:
might be strange for devs working with java11, but our CI will catch it anyway

### DIFF
--- a/build-logic/jvm/src/main/kotlin/build-logic.errorprone.gradle.kts
+++ b/build-logic/jvm/src/main/kotlin/build-logic.errorprone.gradle.kts
@@ -4,11 +4,11 @@ plugins {
     java
 }
 
-if (!project.hasProperty("skipErrorprone")) {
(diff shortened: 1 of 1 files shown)


vlsi:
I thought we use toolchains for the build, so the build should always use Java 17 no matter which Java is the default. Isn't it the case?

vlsi:
Ah, we don't have toolchains in sigstore-java yet.

loosebazooka:
We also want to keep java11 compat, we have all our tests run in Java 11 as well on CI.

vlsi:
We'd better use Java 17 for the build, and use `--release 11` to target Java 11 bytecode/classlib.
Then we could use whatever version for executing the tests.

Using Java 17 for the build would avoid us running into javac11 bugs, and it would make all the devs produce pretty-much the same bytecode.

vlsi:
I suggest we should [require Java 17 for building sigstore-java](https://github.com/sigstore/sigstore-java/pull/923). In that case, this pr could be closed.

vlsi:
I guess this is no longer needed as https://github.com/sigstore/sigstore-java/pull/923 is merged

PR #839, opened 2024-11-01
Update tuf updater api surface

loosebazooka:
- Expose updateMeta and downloadTarget for conformance
- Also don't persist derived attributes in TufMeta

### DIFF
--- a/sigstore-java/src/main/java/dev/sigstore/tuf/Updater.java
+++ b/sigstore-java/src/main/java/dev/sigstore/tuf/Updater.java
@@ -22,6 +22,7 @@ import com.google.common.hash.Hashing;
 import dev.sigstore.encryption.Keys;
 import dev.sigstore.encryption.signers.Verifiers;
 import dev.sigstore.tuf.model.*;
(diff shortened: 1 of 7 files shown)


sigstore-java/src/main/java/dev/sigstore/tuf/Updater.java
```diff
@@ -304,7 +317,6 @@ void updateTimestamp()
             localTimestamp.getSignedMeta().getVersion(), timestamp.getSignedMeta().getVersion());
       }
       if (localTimestamp.getSignedMeta().getVersion() == timestamp.getSignedMeta().getVersion()) {
-        trustedMetaStore.setTimestamp(localTimestamp);
```
patflynn, line 307: I wonder why I was doing this...

patflynn (approved):

sigstore-java/src/main/java/dev/sigstore/tuf/Updater.java
loosebazooka: I think I did this, in some weird refactor.

PR #840, opened 2024-11-04
More tuf updates for conformance

loosebazooka:
- don't persist N.root.json, could clash with a delegation
- validate at parsetime the type of meta file
- generify metastore a bit more, force updater to clear the right meta
- validate any bootstrapped root at update start time
- ~~perform an update on meta if download target is called directly and we haven't updated yet~~

more issues found during conformance, maybe I should just merge the conformance test suite in and fix these as I go from now on.

### DIFF
--- a/sigstore-java/src/main/java/dev/sigstore/tuf/MetaStore.java
+++ b/sigstore-java/src/main/java/dev/sigstore/tuf/MetaStore.java
@@ -15,7 +15,6 @@
  */
 package dev.sigstore.tuf;
 
-import dev.sigstore.tuf.model.Root;
 import dev.sigstore.tuf.model.SignedTufMeta;
 import dev.sigstore.tuf.model.TufMeta;
 import java.io.IOException;
@@ -31,8 +30,7 @@ public interface MetaStore extends MetaReader {
   String getIdentifier();
 
   /**
-   * Generic method to store one of the {@link SignedTufMeta} resources in the local tuf store. Do
-   * not use for Root role, use {@link #writeRoot(Root)} instead.
+   * Generic method to store one of the {@link SignedTufMeta} resources in the local tuf store.
    *
    * @param roleName the name of the role
    * @param meta the metadata to store
@@ -41,26 +39,13 @@ public interface MetaStore extends MetaReader {
   void writeMeta(String roleName, SignedTufMeta<? extends TufMeta> meta) throws IOException;
 
   /**
-   * Once you have ascertained that your root is trustworthy use this method to persist it to your
-   * local store. This will usually only be called with a root loaded statically from a bundled
-   * trusted root, or after the successful verification of an updated root from a mirror.
-   *
-   * @param root a root that has been proven trustworthy by the client
-   * @throws IOException since some implementations may persist the root to disk or over the network
-   *     we throw {@code IOException} in case of IO error.
-   * @see <a
-   *     href="https://theupdateframework.github.io/specification/latest/#detailed-client-workflow">5.3.8</a>
-   */
-  void writeRoot(Root root) throws IOException;
-
-  /**
-   * This clears out the snapshot and timestamp metadata from the store, as required when snapshot
-   * or timestamp verification keys have changed as a result of a root update.
+   * Generic method to remove meta, useful when keys rotated in root. Deletion is not optional,
+   * implementers must ensure meta is removed from the storage medium.
    *
    * @throws IOException implementations that read/write IO to clear the data may throw {@code
    *     IOException}
    * @see <a
    *     href="https://theupdateframework.github.io/specification/latest/#detailed-client-workflow">5.3.11</a>
    */
-  void clearMetaDueToKeyRotation() throws IOException;
+  void clearMeta(String role) throws IOException;
 }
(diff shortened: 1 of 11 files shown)


sigstore-java/src/main/java/dev/sigstore/tuf/MetaStore.java
```diff
@@ -41,26 +39,13 @@ public interface MetaStore extends MetaReader {
...
+   * Generic method to remove meta, useful when keys rotated in root. Deletion is not optional,
+   * implementers must ensure meta is removed from the storage medium.
    *
    * @throws IOException implementations that read/write IO to clear the data may throw {@code
    *     IOException}
    * @see <a
    *     href="https://theupdateframework.github.io/specification/latest/#detailed-client-workflow">5.3.11</a>
    */
-  void clearMetaDueToKeyRotation() throws IOException;
+  void clearMeta(String role) throws IOException;
```
patflynn, line 50: I was following the spec here: https://github.com/theupdateframework/specification/blob/master/tuf-spec.md?plain=1#L1366

Why is this changing?

sigstore-java/src/main/java/dev/sigstore/tuf/Updater.java
```diff
@@ -104,20 +105,24 @@ public void updateMeta() throws IOException, NoSuchAlgorithmException, InvalidKe
...
-    updateTargets();
+    if (!Objects.equals(oldTimestamp.orElse(null), trustedMetaStore.getTimestamp())
+        || trustedMetaStore.findSnapshot().isEmpty()
+        || trustedMetaStore.findTargets().isEmpty()) {
+      // if we need to update or we can't find targets/snapshots locally then grab new snapshot and
+      // targets from remote
+      updateSnapshot();
+      updateTargets();
+    }
+    metaUpdated = true;
```
patflynn, line 116: I'm a bit confused by what this is tracking?  why would it be class level?

sigstore-java/src/main/java/dev/sigstore/tuf/MetaStore.java
loosebazooka: it's not, the storage engine is just being directed by the Updater. So Updater is now telling it what to clear.  We were forcing some logic on the storage engine before?

loosebazooka: https://github.com/sigstore/sigstore-java/pull/840/files#diff-7e790d13f39bb712cf8caac24dda4082ab6b8acb38693cb900a80d268f918b6eR120

sigstore-java/src/main/java/dev/sigstore/tuf/Updater.java
loosebazooka: Yeah this part is strange, but basically a call to "downloadTargets" should call an update if within the context of this updater an update has never occurred. (tuf conformance expects it. Also the go/python clients work this way)

patflynn: I don't remember, but I don't think I designed the Updater to be necessarily single-shot. Even if it was this is kind of ugly for me, but it's difficult for me to understand what the tuf conformance has to do with it.

loosebazooka: it wasn't designed to be single shot. I can instead move this logic to the conformance wrapper. Lemme do that. We can revisit this later.

loosebazooka:
oops hold on

sigstore-java/src/main/java/dev/sigstore/tuf/MetaStore.java
patflynn: oh I see.  yeah that makes sense

sigstore-java/src/test/java/dev/sigstore/tuf/FileSystemTufStoreTest.java
```diff
@@ -43,32 +43,24 @@ void newFileSystemStore_hasRepo(@TempDir Path repoBase) throws IOException {
...
-    TestResources.setupRepoFiles(PROD_REPO, repoBase, "root.json");
-    FileSystemTufStore tufStore = FileSystemTufStore.newFileSystemStore(repoBase);
-    int version = tufStore.readMeta(RootRole.ROOT, Root.class).get().getSignedMeta().getVersion();
-    assertFalse(repoBase.resolve(version + ".root.json").toFile().exists());
-    tufStore.writeRoot(TestResources.loadRoot(TestResources.UPDATER_REAL_TRUSTED_ROOT));
-    assertTrue(repoBase.resolve(version + ".root.json").toFile().exists());
-  }
-
   @Test
   void clearMetaDueToKeyRotation(@TempDir Path repoBase) throws IOException {
```
patflynn, line 57: I think we can remove this test now

loosebazooka: oh yeah perhaps it should be more generic clear, I'll just change that.

patflynn (approved):

PR #852, opened 2024-11-18
Fix test_duplicate_sig_keyids

loosebazooka:
This isn't a security concern, our looping didn't actually allow re-use of keys. But we weren't exactly according to spec.


sigstore-java/src/main/java/dev/sigstore/tuf/Updater.java
```diff
@@ -234,16 +235,23 @@ void verifyDelegate(
       Map<String, Key> publicKeys,
```
patflynn, line 235: github really doesn't want me commenting on non changed lines or something, but line 230 @throws is I guess  out of date?

patflynn (dismissed):

patflynn (approved):

patflynn (approved):

PR #853, opened 2024-11-18
Handle targets with path elements

loosebazooka:
filesystem store stores them down as urlencoded, this mirrors behavior of the go-tuf client


patflynn (approved):
is there no downloadTarget test case?

loosebazooka:
that's handled by conformance, but I'll add it in.

PR #855, opened 2024-11-21
Add support for verifying dsse-intoto

loosebazooka:
- Verification should be able to correctly validate a bundle as cryptographically valid (VerificationOptions.empty())
- Verifiers may also include signer identity during verification
- Verifiers should extract the embedded attestation to do further analysis on the attestation. Sigstore-java does not process those in any way
- There is no signing options for DSSE bundles

needs #873 #872


loosebazooka:
I'm gonna split this up, make it a little easier to review.

sigstore-java/src/main/java/dev/sigstore/KeylessVerifier.java
```diff
@@ -255,4 +262,110 @@ void checkMessageSignature(
...
+    // find one sha256 hash in the subject list that matches the artifact hash
+    if (payload.getSubject().stream()
+        .noneMatch(
+            subject -> {
+              if (subject.getDigest().containsKey("sha256")) {
+                try {
+                  var digestBytes = Hex.decode(subject.getDigest().get("sha256"));
+                  return Arrays.equals(artifactDigest, digestBytes);
+                } catch (DecoderException de) {
+                  // ignore (assume false)
```
patflynn, line 294: log.warn?

patflynn (approved):

sigstore-java/src/main/java/dev/sigstore/KeylessVerifier.java
loosebazooka: I'm not super sure, if it fails then we'll see a keyless verification exception and users can debug at that point? I guess I can add one in a followup.

PR #904, opened 2025-02-22
Update conformance.yml

loosebazooka:
Only run cpython tests on "java17 + prod" invocation

Alleviate the problem a little

### DIFF
--- a/.github/workflows/conformance.yml
+++ b/.github/workflows/conformance.yml
@@ -42,3 +42,4 @@ jobs:
           entrypoint: ${{ github.workspace }}/bin/sigstore-cli
           environment: ${{ matrix.sigstore-env }}
           xfail: "test_verify_dsse_bundle_with_trust_root"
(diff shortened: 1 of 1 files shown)


loosebazooka:
I think my conditional should be inversed

jku (approved):
> Alleviate the problem a little

The problem is that workflow takes forever to run? or is there still an issue with the token availability?

In any case, seems fine to me.
 
Another improvement might be to tweak `test_verify_cpython_release_bundles()` in tuf-conformance: I'm pretty sure it currently verifies every single artifact in every release -- I'm pretty sure this is just wasteful and we could verify one artifact per release instead. This might remove 90% of the processing here. I 'll file an issue

loosebazooka:
yeah that looks good to me, I'll close this.

> The problem is that workflow takes forever to run? or is there still an issue with the token availability?
No, it's mostly that we run conformance like 8 times per PR, which is a bit overkill.

Issue #905, opened 2025-02-23
Documentation: .sigstore.json instead of .sigstore.java in README.md?


thomasleplus:
This [README.md](sigstore-maven-plugin/README.md) talks about avoiding checksum files for `.sigstore.java` files. Isn't it `.sigstore.json`?

Cheers,

Tom

loosebazooka:
Feel free to open a pr or I can fix it when I'm back at a computer

thomasleplus:
Thanks, I've submitted the PR. One more question, the same README.md seems to imply that you don't need to set `-Daether.checksums.omitChecksumsForExtensions=.asc,.sigstore.java` if you use Maven 3.9.2 or higher. But the latest (Maven Artifact Resolver documentation)[https://maven.apache.org/resolver/configuration.html] says that the default value for `aether.checksums.omitChecksumsForExtensions` is `.asc,.sigstore`. So overriding it to `.asc,.sigstore.json` might still be required but I am not sure how to confirm it.

loosebazooka:
Thanks, that other part is a bit confusing. Maven Artifact Resolver sets the default here: https://github.com/apache/maven-resolver/blob/6cd7a95ae1fc868d3f0882fb8bd60c15a87789dd/maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/DefaultArtifactPredicateFactory.java#L54 which probably needs to be updated to: `.asc, .sigstore.json` like this PR: https://github.com/apache/maven-gpg-plugin/pull/109. I'm not sure what this bit is about though: https://github.com/apache/maven-resolver/blob/6cd7a95ae1fc868d3f0882fb8bd60c15a87789dd/maven-resolver-generator-sigstore/src/main/java/org/eclipse/aether/generator/sigstore/SigstoreSignatureArtifactGenerator.java#L46

thomasleplus:
Confusing indeed.

thomasleplus:
I made my first release to Maven Central including the sigstore signatures :+1: But I can confirm that with Maven 3.9.9 create `.sigstore.json.asc` files so the Maven Artifact Resolver needs updating. I can try to track down the corresponding issue tracker and open a ticket or PR their unless you guys already have a channel to the Apache Foundation?

loosebazooka:
I do not. Please go ahead. Happy to help if you run into any issues.

thomasleplus:
[...]

PR #906, opened 2025-02-24
Doc should say .sigstore.json instead of .sigstore.java

thomasleplus:
#### Summary

The PR is to fix a typo in the documentation.

#### Release Note

NONE

#### Documentation

NONE


thomasleplus:
Closes #905

loosebazooka:
The test failures are related to access to the identity token, so we can ignore those. I'll need to update the tests to use a different token for contributions.

PR #923, opened 2025-03-06
chore: use Gradle Java toolchains for the build and test execution

vlsi:
#### Summary

It decouples the build JVM from the test JVM.
By default we use Java 17 for the build and target with `--release 11`.
Then we vary Java version when running tests.

You could control JDK versions with the following Gradle properties:
* `targetJavaVersion`. Java version for source and target compatibility 
* `jdkBuildVersion`. Defaults to 17. [JDK version](https://docs.gradle.org/8.4/userguide/toolchains.html#sec:consuming) to use for building $projectName. If the value is 0, then the current Java is used.
* `jdkBuildVendor`. [JDK vendor](https://docs.gradle.org/8.4/userguide/toolchains.html#sec:vendors) to use building
* `jdkBuildImplementation`. Vendor-specific [virtual machine implementation](https://docs.gradle.org/8.4/userguide/toolchains.html#selecting_toolchains_by_virtual_machine_implementation) to use building
* `jdkTestVersion`. Defaults to `jdkBuildVersion`. [JDK version](https://docs.gradle.org/8.4/userguide/toolchains.html#sec:consuming) to use for testing.
* `jdkTestVendor`. Defaults to `jdkBuildVendor`. [JDK vendor](https://docs.gradle.org/8.4/userguide/toolchains.html#sec:vendors) to use testing
* `jdkTestImplementation`. Defaults to `jdkBuildImplementation`. Vendor-specific [virtual machine implementation](https://docs.gradle.org/8.4/userguide/toolchains.html#selecting_toolchains_by_virtual_machine_implementation) to use testing

#### Release Note

NONE

#### Documentation

NONE


settings.gradle.kts
```diff
@@ -1,5 +1,9 @@
 rootProject.name = "sigstore-java-root"
 
+if (JavaVersion.current() < JavaVersion.VERSION_17) {
+    throw UnsupportedOperationException("Please use Java 17 or 21 for launching Gradle when building sigstore-java, the current Java is ${JavaVersion.current().majorVersion}")
```
loosebazooka, line 4: We still will generate a java11 bytecode final result though? I'm not sure we can eliminate java 11 support just yet.

loosebazooka (approved):

settings.gradle.kts
vlsi: Sure we generate Java 11-compatible bytecode (check for `targetJavaVersion`)

vlsi: See `release.set(buildParameters.targetJavaVersion)`

.github/workflows/ci.yaml
```diff
@@ -58,7 +61,7 @@ jobs:
 
     # if you need to skip staging due to flakiness use "-PskipStaging"
     - name: Test sigstore-java
-      run: ./gradlew build
+      run: ./gradlew build -Porg.gradle.java.installations.auto-download=false -PjdkTestVersion=${{ matrix.java-version }}
```
vlsi, line 64: @loosebazooka , here's how we execute tests with a given Java version even though we always build with Java 17

loosebazooka:
I think you cam merge when ready without examples. I'm working on fixing examples for external contributors using the [conformance token](https://github.com/sigstore-conformance/extremely-dangerous-public-oidc-beacon).

loosebazooka (approved):

PR #925, opened 2025-03-10
Make token string oidc client available outside of cli

loosebazooka:
@vlsi this is step towards making examples work of fork PRs. Once this is in, we need to add an equivalent gradle oidc thing and then inject the token from [conformance](https://github.com/sigstore-conformance/extremely-dangerous-public-oidc-beacon/blob/current-token/oidc-token.txt) into it. wdyt?

### DIFF
--- a/sigstore-cli/src/main/java/dev/sigstore/cli/Sign.java
+++ b/sigstore-cli/src/main/java/dev/sigstore/cli/Sign.java
@@ -18,6 +18,7 @@ package dev.sigstore.cli;
 import dev.sigstore.KeylessSigner;
 import dev.sigstore.TrustedRootProvider;
 import dev.sigstore.oidc.client.OidcClients;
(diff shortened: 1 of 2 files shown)


loosebazooka:
@vlsi can we make this an optional input for sigstore-gradle ? It will help us fix tests on forked requests?

vlsi:
I don't fully get what you mean by "optional input for sigstore-gradle".

Do you mean Gradle should be aware the actual value can differ across executions?
Do you mean sign task gets "up-to-date" no matter what the actual token shows up during execution?

loosebazooka:
Sorry, no I mean, in https://github.com/sigstore/sigstore-java/blob/main/sigstore-gradle/sigstore-gradle-sign-base-plugin/src/main/kotlin/dev/sigstore/sign/OidcClientExtension.kt

PR #957, opened 2025-05-12
Update google-java-format to 1.24.0

loosebazooka:
1.25.0+ require a minimum java build version of 17, so we can update to latest as we upgrade our build.

### DIFF
--- a/build-logic/jvm/src/main/kotlin/build-logic.java.gradle.kts
+++ b/build-logic/jvm/src/main/kotlin/build-logic.java.gradle.kts
@@ -25,7 +25,7 @@ tasks.configureEach<JavaExec> {
 
 spotless {
     java {
-        googleJavaFormat("1.15.0")
+        googleJavaFormat("1.24.0")
         licenseHeaderFile("$rootDir/config/licenseHeader")
         // Note if submodule needs to add more exclusions, it should list ALL of them since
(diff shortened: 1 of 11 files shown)


aaronlew02 (approved):
LGTM

vlsi (approved):

vlsi:
What do you mean by `1.25.0+ require a minimum java build version of 17`?

We currently use Java 17 for the build (still target 11).
We could even bump to "require Java 21 for the build" and still target `--release 11`.

loosebazooka:
I thought we still build with 11 in our matrix

vlsi:
No no no. We use the latest and greatest `javac` compiler, and we use `--release 11` so it produces Java 11-compatible bytecode. This should reduce the number of bugs we face when we use old `javac`.

See https://github.com/sigstore/sigstore-java/pull/923

In practice, you could launch the build with `./gradlew -PjdkBuildVersion=21 jar` to build the jars with `javac 21`.

I've moved pgjdbc to use javac 21 recently: https://github.com/pgjdbc/pgjdbc/pull/3612, and it might be time to bump the build JDK for sigstore-java as well.

That's cool, isn't it?

loosebazooka:
oh yeah, I guess I vaguely recall this now

PR #960, opened 2025-05-14
Add timestamp client and verifier

aaronlew02:
Signed-off-by: Aaron Lew <64337293+aaronlew02@users.noreply.github.com>

Closes: #958

#### Summary

Add a `TimestampClient` to request a timestamp from a timestamp authority and a `TimestampVerifier` to verify a timestamp.


sigstore-java/src/main/java/dev/sigstore/timestamp/client/TimestampClient.java
```diff
@@ -0,0 +1,32 @@
...
+import dev.sigstore.rekor.client.RekorResponse;
+import java.io.IOException;
+import org.bouncycastle.tsp.TSPException;
+
+/** A client to communicate with a timestamp service instance. */
+public interface TimestampClient {
+  /**
+   * Request a timestanp for a timestamp authority.
+   *
+   * @param tsReq
```
loosebazooka, line 27: I think we should add a description to this param:
```
@param tsReq a structured request for a timestamp
```

sigstore-java/src/main/java/dev/sigstore/timestamp/client/TimestampClient.java
```diff
@@ -0,0 +1,32 @@
...
+   * @return a {@link RekorResponse} with information about the log entry
+   */
+  TimestampResponse timestamp(TimestampRequest tsReq)
```
loosebazooka, line 30: I would like for this to only throw TimestampException and encapsulate the rest inside that.

sigstore-java/src/main/java/dev/sigstore/timestamp/client/TimestampClientHttp.java
```diff
@@ -0,0 +1,167 @@
...
+/** A client to communicate with a timestamp service instance. */
+public class TimestampClientHttp implements TimestampClient {
+  private static final URI SIGSTORE_TSA_URI =
+      URI.create("https://timestamp.sigstage.dev/api/v1/timestamp");
+  private static final String CONTENT_TYPE_TIMESTAMP_QUERY = "application/timestamp-query";
+  private static final String ACCEPT_TYPE_TIMESTAMP_REPLY = "application/timestamp-reply";
+
+  private static final Map<String, ASN1ObjectIdentifier> HASH_ALGORITHM_OIDS;
+
+  static {
```
loosebazooka, line 49: this can be created on one line kinda like:

```
    private static final ImmutableMap<String, ASN1ObjectIdentifier> HASH_ALGORITHM_OIDS =
            ImmutableMap.of(
                    "SHA256", TSPAlgorithms.SHA256,
                    "SHA384", TSPAlgorithms.SHA384,
                    "SHA512", TSPAlgorithms.SHA512
            );
```

loosebazooka: alternatively, we could add an Enum to TimestampRequest that strictly defines these supported hashes, and we wouldn't have to check them later.

sigstore-java/src/main/java/dev/sigstore/timestamp/client/TimestampClientHttp.java
```diff
@@ -0,0 +1,167 @@
...
+  public TimestampResponse timestamp(TimestampRequest tsReq) throws TimestampException {
+    TimeStampRequestGenerator bcTsReqGen = new TimeStampRequestGenerator();
+
+    // Map the hash algorithm string from the request to a Bouncy Castle OID
+    var reqAlgorithm = tsReq.getHashAlgorithm();
+    var bcAlgorithmOid = HASH_ALGORITHM_OIDS.get(reqAlgorithm.toUpperCase(Locale.ROOT));
+    if (bcAlgorithmOid == null) {
+      throw new IllegalArgumentException(
+          "Unsupported hash algorithm for TSA request: " + reqAlgorithm);
+    }
```
loosebazooka, line 104: We could avoid this with the Enum

sigstore-java/src/main/java/dev/sigstore/timestamp/client/TimestampRequest.java
```diff
@@ -0,0 +1,35 @@
...
+import java.math.BigInteger;
+import java.util.Optional;
+import org.immutables.value.Value.Immutable;
+
+@Immutable
+public interface TimestampRequest {
+  /** The hash algorithm used to hash the artifact. */
+  String getHashAlgorithm();
+
+  /** The hash of the artifact to be timestamped. */
```
loosebazooka, line 27: I would add in this comment, that for sigstore-java, we want to hash the signature, not the artifact in a signing event.

sigstore-java/src/main/java/dev/sigstore/timestamp/client/TimestampRequest.java
```diff
@@ -0,0 +1,35 @@
...
+  byte[] getHash();
+
+  /** A nonce to prevent replay attacks. */
+  BigInteger getNonce();
+
+  /** Whether or not to include certificates in the response. */
+  Optional<Boolean> getRequestCertificates();
```
loosebazooka, line 34: I think this can just be `boolean requestCertificates();` maybe? default is false. I don't know that `Optional` is adding any value.

sigstore-java/src/test/java/dev/sigstore/timestamp/client/TimestampClientHttpTest.java
```diff
@@ -0,0 +1,248 @@
...
+import org.bouncycastle.tsp.TSPAlgorithms;
+import org.bouncycastle.tsp.TSPException;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
+
+public class TimestampClientHttpTest {
+  private static HttpRequestFactory requestFactory;
+  private static HttpParams httpParams;
+  private static TimestampRequest tsReq;
+  private static final Map<String, ASN1ObjectIdentifier> HASH_ALGORITHM_OIDS;
```
loosebazooka, line 54: To avoid repeating this here, just make HASH_ALGORITHM_IDS in `TimestampClientHttp` package private and access it here.

sigstore-java/src/main/java/dev/sigstore/timestamp/client/TimestampVerifier.java
```diff
@@ -0,0 +1,273 @@
...
+    pkixParams.setDate(tsDate);
+
+    cpv.validate(tsa.getCertPath(), pkixParams);
+  }
+
+  /**
+   * Finds the TSA that verifies the provided timestamp token by validating its signature using the
+   * certificate chain embedded within the token and matching the leaf to a known TSA.
+   */
+  CertificateAuthority findVerifyingTsaFromEmbeddedCerts(TimeStampToken tsToken)
```
haydentherapper, line 195: Flagging this for more discussion: I wonder if we should drop support for verifying timestamps using embedded certificates, and similarly on the write path, drop support for setting the `certReq` bit.

Embedded certificates are an alternative to distributing a full PKI chain, but you do still need a root of trust. Currently, the TUF metadata assumes you've specified the full chain - you could technically specify just a root, but in practice, a deployment operator should just specify the full chain if they've already specified a root.

cc @loosebazooka

sigstore-java/src/main/java/dev/sigstore/timestamp/client/TimestampRequest.java
```diff
@@ -0,0 +1,35 @@
...
+@Immutable
+public interface TimestampRequest {
+  /** The hash algorithm used to hash the artifact. */
+  String getHashAlgorithm();
+
+  /** The hash of the artifact to be timestamped. */
+  byte[] getHash();
+
+  /** A nonce to prevent replay attacks. */
+  BigInteger getNonce();
```
loosebazooka, line 31: we might want to just generate this by default, and not require the user to provide it?

sigstore-java/src/test/java/dev/sigstore/timestamp/client/TimestampClientHttpTest.java
```diff
@@ -0,0 +1,248 @@
...
+        ImmutableTimestampRequest.builder()
+            .hash(artifactHashBytes)
+            .hashAlgorithm("SHA256")
+            .nonce(nonce)
+            .build();
+  }
+
+  @Test
+  public void timestamp_success()
+      throws TimestampException, IOException, TSPException, CertificateException {
```
loosebazooka, line 88: you can just group all these exceptions as `Exception` in tests.

sigstore-java/src/test/java/dev/sigstore/timestamp/client/TimestampVerifierTest.java
```diff
@@ -0,0 +1,258 @@
...
+  public static void loadResources() throws IOException {
+    // Response from Sigstore TSA (in trusted root) with embedded certs
+    try (var is =
+        Resources.getResource(
+                "dev/sigstore/samples/timestamp-response/valid/sigstore_tsa_response_with_embedded_certs.tsr")
+            .openStream()) {
+      if (is == null) {
+        throw new IOException(
+            "dev/sigstore/samples/timestamp-response/valid/sigstore_tsa_response_with_embedded_certs.tsr");
+      }
```
loosebazooka, line 59: Handling `is == null` is maybe too cautious, just let the init fail in tests.

loosebazooka (commented):
Just some preliminary comments, I'll go over the logic in TimestampClientHttp on a second pass.

sigstore-java/src/main/java/dev/sigstore/timestamp/client/TimestampVerifier.java
```diff
@@ -0,0 +1,273 @@
...
+          NoSuchAlgorithmException {
+    return newTimestampVerifier(trustedRoot.getTSAs());
+  }
+
+  public static TimestampVerifier newTimestampVerifier(List<CertificateAuthority> tsas)
+      throws InvalidKeySpecException,
+          NoSuchAlgorithmException,
+          InvalidAlgorithmParameterException,
+          CertificateException {
+    // check to see if we can use all TSAs (this is a bit eager)
```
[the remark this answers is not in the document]
loosebazooka: That's validating that tsas can be converted into a trust anchor at construction time. We do something similar in the fulcio verifier.

sigstore-java/src/main/java/dev/sigstore/timestamp/client/TimestampVerifier.java
loosebazooka: Yeah I'm not sure, we kinda landed on this so we could handle all use-cases. But if clients all agree to ignore, I'm fine with that.

sigstore-java/src/main/java/dev/sigstore/timestamp/client/TimestampVerifier.java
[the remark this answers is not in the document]
aaronlew02: This is a duplicate of [this comment](https://github.com/sigstore/sigstore-java/blob/dd78f82524a1dcedbd2679840cd6990fbe314926/sigstore-java/src/main/java/dev/sigstore/fulcio/client/FulcioVerifier.java#L81), which suggests that this approach is very proactive since it validates all TSAs at initialization.

@loosebazooka (who suggested keeping it here as well) can elaborate if necessary.

sigstore-java/src/main/java/dev/sigstore/timestamp/client/TimestampVerifier.java
haydentherapper: I've reminded myself what `certReq` actually does, it's for embedding just the TSA leaf certificate in the TimestampResponse. I don't think it's a bad thing to keep this logic in to cover more cases, but if we do, we'll need to change how we find the correct TSA chain in the trust root. Currently, it looks like it's looking for a matching leaf certificate in the TSA chain in the trust root, but I wouldn't expect a chain to contain the leaf cert if it's included in the TSR already. Instead, we'd have to try to verify the full chain, using the CA chain from the trust root and the leaf cert from the TSR.

sigstore-java/src/main/java/dev/sigstore/timestamp/client/TimestampClientHttp.java
```diff
@@ -0,0 +1,167 @@
...
+import org.bouncycastle.asn1.ASN1ObjectIdentifier;
+import org.bouncycastle.tsp.TSPAlgorithms;
+import org.bouncycastle.tsp.TSPException;
+import org.bouncycastle.tsp.TimeStampRequest;
+import org.bouncycastle.tsp.TimeStampRequestGenerator;
+import org.bouncycastle.tsp.TimeStampResponse;
+
+/** A client to communicate with a timestamp service instance. */
+public class TimestampClientHttp implements TimestampClient {
+  private static final URI SIGSTORE_TSA_URI =
```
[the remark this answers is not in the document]
aaronlew02: Currently, only the staging TSA supports disallowing hash digest lengths inconsistent with the specified hash algorithm.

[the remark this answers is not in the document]
haydentherapper: Ah, we'll get prod updated.

sigstore-java/src/main/java/dev/sigstore/timestamp/client/TimestampVerifier.java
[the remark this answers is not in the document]
loosebazooka: It's mostly to hit a configuration error early (to prevent a weird failure later where calling the verifier)

sigstore-java/src/main/java/dev/sigstore/timestamp/client/TimestampClientHttp.java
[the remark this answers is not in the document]
loosebazooka: We hardcode prod and staging on all clients. But that'll change with signing config. I think this is fine for now.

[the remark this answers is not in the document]
aaronlew02: I'm hardcoding the staging URI exclusively in the client due to the digest length validation functionality. In all other contexts (i.e., the trusted root and the verifier tests), the prod URI is used.

sigstore-java/src/main/java/dev/sigstore/timestamp/client/TimestampVerifier.java
loosebazooka: I think this matching is okay?
1. Extract the leaf
2. Match it again a chain in the trusted_root
3. Check the signature
4. Ensure the found TSA has a consistent chain (in validateTsaChain)

haydentherapper: Rephrasing slightly:

1. Either a) extract the leaf from the Timestamp Response, or b) extract the leaf from the chain in the trusted root
2. Use only the CA certificates in the trusted root to build a pool of intermediate and root certs (sorry, not sure how it's done in java, this is based on Go)
3. Verify the chain from leaf to root (validateTSAChain I think?)
4. Using the public key from the leaf cert, check the signature

loosebazooka: Sorry yeah, this *is* what we're doing here. I gotta double check the chain verification.
There isn't really a unique in tsa land though. TSA is just providing its certificate, the same leaf will be embedded in all timestamps if requested. This leaf is also provided as part of the chain in our trusted_root.

haydentherapper: > This leaf is also provided as part of the chain in our trusted_root

This is the part that I was wondering if we should assume. Would we be OK with a leaf certificate not in the trusted root chain?

loosebazooka: I didn't think so. The leaf is what identifies the TSA, isn't it?

haydentherapper: Ah I see, you're viewing the embedded leaf as a link to find the correct chain in the trusted root, yea?

I think this is just a different viewpoint, whether you try to verify against every chain in the trust root ([what sigstore-go is doing](https://github.com/sigstore/sigstore-go/blob/cedac1b7e7d4ff3170c12f24f6b53d49a6396c51/pkg/verify/tsa.go#L98) or whether you can find the specific chain with the leaf.

OK, I think both are reasonable approaches.

loosebazooka: I mean I guess we didn't have to verify the chain of it comes from the trust root.

sigstore-java/src/main/java/dev/sigstore/timestamp/client/TimestampResponse.java
```diff
@@ -0,0 +1,23 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package dev.sigstore.timestamp.client;
+
+import org.immutables.value.Value.Immutable;
+
+@Immutable
+public interface TimestampResponse {
+  byte[] getBytes();
```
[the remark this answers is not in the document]
aaronlew02: `getBytes`, as opposed to the two names you suggested, places an emphasis on the returned content being the byte encoding that is wrapped in and passed between our and Bouncy Castle's response objects. In other words, it may be best to avoid having a field identical to the name of the object itself.

Bouncy Castle uses [`getEncoded`](https://downloads.bouncycastle.org/java/docs/bcpkix-jdk13-javadoc/org/bouncycastle/tsp/TimeStampResponse.html#getEncoded()), which we can use for uniformity. Alternatively, we could use `getResponseBytes` or something similar.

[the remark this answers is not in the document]
loosebazooka: `getEncoded` works. `getTSR` or `getTimestampResponse` feel redundant to me.

sigstore-java/src/main/java/dev/sigstore/timestamp/client/TimestampVerifier.java
haydentherapper: Aaron and I chatted about this - the primary difference between what I was asking about and what's implemented here is just ordering, and I've realized that there is not a duplicated chain verification anywhere. There's no actual risk with doing this in a different order than what I mentioned as well, so this LGTM!

loosebazooka (commented):
looks good, just some minor things

.gitignore
```diff
@@ -31,3 +31,6 @@
 # except this icon
 !/.idea/icon.png
 
+# vscode java output directories
+/bin/=
```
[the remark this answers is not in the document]
aaronlew02: Not sure how that got there. Removed.

loosebazooka (approved):
nice.

PR #964, opened 2025-05-16
chore: migrate to Kotlin Dokka 2.0

vlsi:
#### Summary

Dokka 2 is still experimental, however, we use very basic apis, so let's migrate and silence the warning.

#### Release Note
NONE

#### Documentation
NONE


loosebazooka (approved):

vlsi:
By the way, I've no idea if we should keep publishing the documentation in "javadoc" format rather than html.

dokkaHtml feels better for Kotlin-based projects. For instance, dokka html format would probably describe extension functions better than javadoc.

At the same time, I've absolutely no clue which tools use `-javadoc.jar` artifacts from Central :)

loosebazooka:
I guess we don't publish the plugin to central so it doesn't matter? I'll just merge this for now to remove warning and we can revisit later.

vlsi:
> I guess we don't publish the plugin to central so it doesn't matter?

It doesn't hurt publishing it.
I typically publish my Gradle plugins to both Central and Gradle's Plugin Portal for redundancy.

However, let's migrate to html later if needed. ChatGPT suggests one should use "javadoc" when publishing dokka-generated artifacts to Central :)

PR #965, opened 2025-05-19
Consume signing_config v0.2 from TUF repo if availalbe.

loosebazooka:
#### Summary
This has the sigstore TUF client attempt to download and parse the signing config, may return null if not found.


aaronlew02 (approved):
LGTM

haydentherapper (approved):

PR #966, opened 2025-05-22
ignoreUnknownFields when parsing json

loosebazooka:
#### Summary
As fields are changed an populated in our protos, make sure we don't fail on a non-breaking addition to the specs.

Should fix our staging failure for now, but we need to make sure we update our high value clients and we move forward.


haydentherapper (approved):
Rubber stamping

aaronlew02 (approved):

PR #967, opened 2025-05-23
Add providers for signing config and legacy helper

loosebazooka:
And some other minor associated changes as part #954

### DIFF
--- a/sigstore-java/src/main/java/dev/sigstore/SigningConfigProvider.java
+++ b/sigstore-java/src/main/java/dev/sigstore/SigningConfigProvider.java
@@ -1,0 +1,74 @@
+/*
+ * Copyright 2025 The Sigstore Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package dev.sigstore;
+
+import com.google.common.base.Preconditions;
+import dev.sigstore.trustroot.SigstoreConfigurationException;
+import dev.sigstore.trustroot.SigstoreSigningConfig;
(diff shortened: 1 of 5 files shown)


sigstore-java/src/main/java/dev/sigstore/SigningConfigProvider.java
```diff
@@ -0,0 +1,74 @@
+/*
+ * Copyright 2023 The Sigstore Authors.
```
aaronlew02, line 2: nit: copyright year

sigstore-java/src/main/java/dev/sigstore/SigningConfigProvider.java
```diff
@@ -0,0 +1,74 @@
...
+      SigstoreTufClient.Builder tufClientBuilder, SigstoreSigningConfig defaultConfig) {
+    Preconditions.checkNotNull(tufClientBuilder);
+    return () -> {
+      try {
+        var tufClient = tufClientBuilder.build();
+        tufClient.update();
+        var fromTuf = tufClient.getSigstoreSigningConfig();
+        return fromTuf == null ? defaultConfig : fromTuf;
+      } catch (IOException ex) {
+        throw new SigstoreConfigurationException(
```
haydentherapper, line 57: You could return `defaultConfig` if not null on error?

sigstore-java/src/main/java/dev/sigstore/trustroot/LegacySigningConfig.java
```diff
@@ -0,0 +1,58 @@
...
+  static final URI REKOR_PUBLIC_GOOD_URI = URI.create("https://rekor.sigstore.dev");
+  static final URI REKOR_STAGING_URI = URI.create("https://rekor.sigstage.dev");
+
+  static final URI FULCIO_PUBLIC_GOOD_URI = URI.create("https://fulcio.sigstore.dev");
+  static final URI FULCIO_STAGING_URI = URI.create("https://fulcio.sigstage.dev");
+
+  static final URI DEX_PUBLIC_GOOD_URI = URI.create("https://oauth2.sigstore.dev/auth");
+  static final URI DEX_STAGING_GOOD_URI = URI.create("https://oauth2.sigstage.dev/auth");
+
+  static final URI TSA_PUBLIC_GOOD_URI = URI.create("https://tsa.sigstore.dev");
```
haydentherapper, line 37: For the TSA, can this be the full path rather than just the host? The TSA client shouldn't assume that the path is the same for all TSAs, so that callers can use other TSAs besides the sigstore/timestamp-authority implementation.

PR #975, opened 2025-06-02
Use Sigstore staging TSA in timestamp client

aaronlew02:
Partially addresses #959 

#### Summary

This change sets up the timestamp client to use staging URIs for the Sigstore TSA, as timestamping functionality is currently restricted to staging environments.


loosebazooka:
ignore those failing windows example tests.

PR #977, opened 2025-06-02
Add artifact validation and staging tests to timestamp verifier

aaronlew02:
Partially addresses #959

#### Summary

This changes enables artifact validation (i.e., confirmation that the message imprint digest of a timestamp matches the artifact digest sent in the request) in the timestamp verifier. 
This change also updates the verifier's tests to include staging URIs for the Sigstore TSA, as timestamping functionality is currently restricted to staging environments.


sigstore-java/src/main/java/dev/sigstore/timestamp/client/TimestampVerifier.java
```diff
@@ -133,16 +137,36 @@ public void verify(TimestampResponse tsResp) throws TimestampVerificationExcepti
...
+          artifactDigest = Hashing.sha384().hashBytes(artifact).asBytes();
+          break;
+        case SHA512:
+          artifactDigest = Hashing.sha512().hashBytes(artifact).asBytes();
+          break;
+        default:
+          throw new IllegalStateException(); // We shouldn't be here.
+      }
+      validateTokenMessageImprintDigest(tsToken, artifactDigest);
+    } catch (TimestampException | TimestampVerificationException tsve) {
```
loosebazooka, line 167: I can't tell where `TimestampException` comes from here.

Wrapping `TimestampVerificationException` in another of the same is a bit odd. I think you can move the code in `validateTokenMessageImprintDigest` just into this method directly and throw that exception.

aaronlew02: Keep everything else as is?

loosebazooka: I think you can let the exception from `validateTokenMessageImprintDigest` passthrough without catching and rethrowing it

PR #979, opened 2025-06-04
Add RFC3161 timestamps to keyless signer and verifier

### LINKED ISSUE #699: Support validation based on rfc3161 from TSA (vs SET from rekor)
https://github.com/sigstore/protobuf-specs/blob/main/protos/sigstore_bundle.proto#L108

https://github.com/sigstore/protobuf-specs/blob/main/protos/sigstore_common.proto#L141

### LINKED ISSUE #959: Add timestamp to bundle and validate

aaronlew02:
Closes #959 and thus closes #699

#### Summary

This change allows for the `KeylessSigner` to add RFC3161 timestamps to a bundle and `KeylessVerifier` to verify such timestamps.

### DIFF
--- a/sigstore-java/src/main/java/dev/sigstore/KeylessSigner.java
+++ b/sigstore-java/src/main/java/dev/sigstore/KeylessSigner.java
@@ -26,6 +26,7 @@ import dev.sigstore.bundle.Bundle;
 import dev.sigstore.bundle.Bundle.HashAlgorithm;
 import dev.sigstore.bundle.Bundle.MessageSignature;
 import dev.sigstore.bundle.ImmutableBundle;
+import dev.sigstore.bundle.ImmutableTimestamp;
 import dev.sigstore.encryption.certificates.Certificates;
 import dev.sigstore.encryption.signers.Signer;
 import dev.sigstore.encryption.signers.Signers;
@@ -46,6 +47,13 @@ import dev.sigstore.rekor.client.RekorParseException;
 import dev.sigstore.rekor.client.RekorResponse;
 import dev.sigstore.rekor.client.RekorVerificationException;
 import dev.sigstore.rekor.client.RekorVerifier;
+import dev.sigstore.timestamp.client.ImmutableTimestampRequest;
+import dev.sigstore.timestamp.client.TimestampClient;
+import dev.sigstore.timestamp.client.TimestampClientHttp;
+import dev.sigstore.timestamp.client.TimestampException;
+import dev.sigstore.timestamp.client.TimestampResponse;
+import dev.sigstore.timestamp.client.TimestampVerificationException;
+import dev.sigstore.timestamp.client.TimestampVerifier;
 import dev.sigstore.trustroot.SigstoreConfigurationException;
 import dev.sigstore.tuf.SigstoreTufClient;
 import java.io.IOException;
@@ -89,6 +97,8 @@ public class KeylessSigner implements AutoCloseable {
   private final FulcioVerifier fulcioVerifier;
   private final RekorClient rekorClient;
   private final RekorVerifier rekorVerifier;
+  private final TimestampClient timestampClient;
+  private final TimestampVerifier timestampVerifier;
   private final OidcClients oidcClients;
   private final List<OidcTokenMatcher> oidcIdentities;
   private final Signer signer;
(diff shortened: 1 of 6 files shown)


sigstore-java/src/main/java/dev/sigstore/tuf/FileSystemTufStore.java
```diff
@@ -96,8 +97,8 @@ public void writeMeta(String roleName, SignedTufMeta<?> meta) throws IOException
   }
 
   @Override
-  public <T extends SignedTufMeta<?>> Optional<T> readMeta(String roleName, Class<T> tClass)
-      throws IOException {
+  public <T extends SignedTufMeta<? extends TufMeta>> Optional<T> readMeta(
```
loosebazooka, line 100: this seems unrelated

sigstore-java/src/main/java/dev/sigstore/KeylessVerifier.java
```diff
@@ -188,11 +196,55 @@ public void verify(byte[] artifactDigest, Bundle bundle, VerificationOptions opt
...
+      try {
+        TimeStampResponse bcTsResp = new TimeStampResponse(tsBytes);
+        TimeStampToken tsToken = bcTsResp.getTimeStampToken();
+        if (tsToken == null) {
+          throw new KeylessVerificationException(
+              "Timestamp token is missing from timestamp response");
+        }
+        TimeStampTokenInfo tsInfo = tsToken.getTimeStampInfo();
+        if (tsInfo == null) {
+          throw new KeylessVerificationException("Timestamp info is missing from timestamp token");
```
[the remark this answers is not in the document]
aaronlew02: `FulcioVerifier` and `RekorVerifier` do not return anything. Would it be worth breaking precedent to prevent parsing here?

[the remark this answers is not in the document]
loosebazooka: I think I would ideally just like for KeylessVerifier not to be inspecting the bc stuff.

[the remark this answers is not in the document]
aaronlew02: Addressed by adding a static method `getGenTime` to `TimestampResponse`.

loosebazooka (commented):
LGTM, just that minor refactoring.

sigstore-java/src/main/java/dev/sigstore/KeylessSigner.java
```diff
@@ -355,13 +385,43 @@ public List<Bundle> sign(List<byte[]> artifactDigests) throws KeylessSignerExcep
...
+      var bundleBuilder =
           ImmutableBundle.builder()
               .certPath(signingCert)
               .addEntries(rekorResponse.getEntry())
               .messageSignature(
-                  MessageSignature.of(HashAlgorithm.SHA2_256, artifactDigest, signature))
-              .build());
+                  MessageSignature.of(HashAlgorithm.SHA2_256, artifactDigest, signature));
+
+      // Timestamp functionality enabled only in staging
```
loosebazooka, line 395: I think this comment makes a little more sense if we say something like:
```
// Timestamp functionality only enabled if timestampUri is provided
```

sigstore-java/src/main/java/dev/sigstore/KeylessSigner.java
```diff
@@ -355,13 +385,43 @@ public List<Bundle> sign(List<byte[]> artifactDigests) throws KeylessSignerExcep
...
+          tsResp = timestampClient.timestamp(tsReq);
+        } catch (TimestampException ex) {
+          throw new KeylessSignerException("Failed to generate timestamp", ex);
+        }
+
+        try {
+          timestampVerifier.verify(tsResp, signature);
+        } catch (TimestampVerificationException ex) {
+          throw new KeylessSignerException("Returned timestamp was invalid", ex);
+        }
```
loosebazooka, line 416: I wonder if we should check if the timestamp is truly in the validity period of the cert here as well?

loosebazooka: maybe we should just call KeylessVerifier.verify or something instead of doing any manual validation here?

loosebazooka (approved):
cool, lets enable the conformance test that used timestamps now that this is in?

PR #985, opened 2025-06-06
Reduce redundant action runs

loosebazooka:
Looks like we were running a lot of redundant checks here. Match what we do in other sigstore repos.


aaronlew02 (approved):

.github/workflows/ci.yaml
```diff
@@ -11,12 +11,9 @@ name: Test
...
 on:
   push:
-    branches:
-      - '*'
-  pull_request:
-    branches:
-      - main
-  workflow_call: # allow this workflow to be called by other workflows
+    branches: [main]
+  pull_request: {}
```
vlsi: This effectively removed execution of Test workflows for external contributors

PR #1013, opened 2025-07-08
Allow RekorEntry to be built from TransparencyLogEntry

aaronlew02:
Closes #1012 

#### Summary

This change modifies the `RekorEntry` object to allow building from a `TransparencyLogEntry` proto returned by Rekor v2. This will allow for universal use of `RekorEntry`.

PR #1014, opened 2025-07-08
Add Rekor v2 staging URI to LegacySigningConfig

aaronlew02:
#### Summary

This change adds the staging URI for Rekor v2 to the `LegacySigningConfig`.

### DIFF
--- a/sigstore-java/src/main/java/dev/sigstore/trustroot/LegacySigningConfig.java
+++ b/sigstore-java/src/main/java/dev/sigstore/trustroot/LegacySigningConfig.java
@@ -28,6 +28,7 @@ public class LegacySigningConfig {
 
   static final URI REKOR_PUBLIC_GOOD_URI = URI.create("https://rekor.sigstore.dev");
   static final URI REKOR_STAGING_URI = URI.create("https://rekor.sigstage.dev");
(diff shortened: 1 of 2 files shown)


sigstore-java/src/main/java/dev/sigstore/trustroot/LegacySigningConfig.java
```diff
@@ -48,16 +49,25 @@ static SigstoreSigningConfig from(URI fulcioUrl, URI rekorUrl, URI dexUrl, @Null
...
             .addCas(Service.of(fulcioUrl, 1))
-            .addTLogs(Service.of(rekorUrl, 1))
             .addOidcProviders(Service.of(dexUrl, 1));
+
+    if (rekorUrl.equals(REKOR_V2_STAGING_URI)) {
+      signingConfigBuilder.addTLogs(Service.of(rekorUrl, 2));
+    } else {
+      signingConfigBuilder.addTLogs(Service.of(rekorUrl, 1));
+    }
+
```
loosebazooka, line 59: I think we should do this another way. `from(...)` is meant to be generic without knowlege of URIs. We can change the inputs to be Service instead of URI?

loosebazooka (commented):
Could this PR can be made independent of #1013? I can't tell if theres any actual overlap.

PR #1016, opened 2025-07-08
Unify Rekor v1 and v2 verifiers

### LINKED ISSUE #1015: Unify Rekor v1 and v2 verifiers
This issue proposes to unify the `RekorVerifier` (used to verify Rekor v1 entries) and `RekorV2Verifier` (used to verify Rekor v2 entries) into a single `RekorVerifier` capable of verifying both versions.

aaronlew02:
Closes #1015 

#### Summary

This change unifies the `RekorVerifier` (used to verify Rekor v1 entries) and `RekorV2Verifier` (used to verify Rekor v2 entries) into a single `RekorVerifier` capable of verifying both versions.


sigstore-java/src/main/java/dev/sigstore/rekor/client/RekorVerifier.java
```diff
@@ -20,20 +20,23 @@
...
 import java.security.spec.InvalidKeySpecException;
+import java.time.Instant;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Base64;
 import java.util.List;
+import java.util.Optional;
 import org.bouncycastle.util.encoders.Hex;
 
-/** Verifier for rekor entries. */
```
loosebazooka, line 36: lets keep the double `/**` <-- implies it will generate javadoc

sigstore-java/src/main/java/dev/sigstore/rekor/client/RekorVerifier.java
```diff
@@ -50,42 +53,45 @@ private RekorVerifier(List<TransparencyLog> tlogs) {
...
+      throw new RekorVerificationException("No inclusion proof in entry.");
     }
 
     var tlog =
-        TransparencyLog.find(tlogs, Hex.decode(entry.getLogID()), entry.getIntegratedTimeInstant())
+        TransparencyLog.find(tlogs, Hex.decode(entry.getLogID()), timestamps)
             .orElseThrow(
                 () ->
                     new RekorVerificationException(
                         "Log entry (logid, timestamp) does not match any provided transparency logs."));
```
loosebazooka, line 76: I think this error message might need some work, but I'm not sure how because now we must consider all timestamps? I'm actually still not sure how exactly we want to handle finding the log.

We need to check some conditions:
1. How many valid timestamps we require to exist within the log's validity period.

sigstore-java/src/main/java/dev/sigstore/trustroot/TransparencyLog.java
```diff
@@ -59,4 +58,21 @@ static Optional<TransparencyLog> find(List<TransparencyLog> all, byte[] logId, I
...
+
+  /**
+   * Find a log by list of validity times and logId. This will find the first log with matching log
+   * id that was valid at all times.
+   *
+   * @param logId the logId of the log
+   * @param times the list of timea the log was expected to be valid
+   * @return the first log that was valid at all {@code times} with logId {@code logId}
+   */
+  static Optional<TransparencyLog> find(
```
loosebazooka, line 70: we run the risk of timestamps that don't really matter affecting how we find a log here and I think we might need to consider the implications of that.

loosebazooka (commented):
looks good, I think we should take another pass as a group on deciding how to handle time.

sigstore-java/src/main/java/dev/sigstore/rekor/client/RekorVerifier.java
loosebazooka: I'm starting to think that if at least one valid timestamp identifies the log then we should be good to go. @haydentherapper we will never have overlapping logids/keys will we?

haydentherapper: Log IDs/checkpoint key IDs should always be unique for each instance. Technically a key could be reused, but for our Rekor v2 deployments, we will never reuse key material.

re: timestamps, I'll take this offline since there's a few threads about this.

loosebazooka: @aaronlew02 lets add a TODO(#<issue number>) here and unblock this review. We can revisit this in the future.

sigstore-java/src/test/java/dev/sigstore/rekor/v2/client/RekorV2ClientHttpTest.java
```diff
@@ -62,9 +62,9 @@ public void putEntry() throws Exception {
     var entry = client.putEntry(req);
 
     assertNotNull(entry);
-    assertNotNull(entry.getInclusionProof());
+    assertNotNull(entry.getVerification().getInclusionProof());
     assertTrue(entry.getLogIndex() >= 0);
-    assertNotNull(entry.getLogId());
+    assertNotNull(entry.getLogID());
```
loosebazooka, line 67: hrmm.. I didn't notice this earlier. The java style guide wants this to be `getLogId` (where is this coming from?)

loosebazooka (approved):

sigstore-java/src/test/java/dev/sigstore/rekor/v2/client/RekorV2ClientHttpTest.java
aaronlew02: This is coming from [RekorEntry](https://github.com/sigstore/sigstore-java/blob/085e06f18a760f9a8963184e15f5e2bc1e412a4f/sigstore-java/src/main/java/dev/sigstore/rekor/client/RekorEntry.java#L163). I agree that the capitalization of `ID` is not ideal.

loosebazooka (approved):
do we have a test with no timestamp and no integrated time?

loosebazooka (approved):

PR #1021, opened 2025-07-16
Move TLogEntry-to-RekorEntry converter to ProtoMutators

aaronlew02:
#### Summary

This changes moves a method that converts a `TransparencyLogEntry` proto object into a `RekorEntry` Java object into the relevant `ProtoMutators` so as to avoid exposing a proto object.


sigstore-java/src/main/java/dev/sigstore/proto/ProtoMutators.java
```diff
@@ -59,4 +68,41 @@ public static HashAlgorithm from(Bundle.HashAlgorithm algorithm) {
     }
     throw new IllegalStateException("Unknown hash algorithm: " + algorithm);
   }
+
+  public static RekorEntry fromTLogEntry(TransparencyLogEntry tle) throws RekorParseException {
```
loosebazooka, line 72: I think I did a bad job with naming in this class. I think more appropriate name would be `toRekorEntry(...)`, I'll go update the other names after this PR.

loosebazooka (dismissed):

PR #1035, opened 2025-07-25
fix: Get Rekor v2 from signing config if enabled

aaronlew02:
#### Summary

This changes fixes an inverse in the condition in `KeylessSigner` that is used to get Rekor v2 from the signing config if enabled.


loosebazooka (approved):
We should probably test this somehow

PR #1037, opened 2025-07-28
fix(test): Verify signing result for Rekor v1 vs. v2

aaronlew02:
#### Summary

This change updates the mechanism for verifying the signing result in `KeylessTest` to check if the API version matches the version of Rekor enabled.

### DIFF
--- a/sigstore-java/src/test/java/dev/sigstore/KeylessTest.java
+++ b/sigstore-java/src/test/java/dev/sigstore/KeylessTest.java
@@ -64,7 +64,7 @@ public class KeylessTest {
     var signer = KeylessSigner.builder().sigstorePublicDefaults().build();
     var results = signer.sign(artifactDigests);
 
-    verifySigningResult(results);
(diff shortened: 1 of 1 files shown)


sigstore-java/src/test/java/dev/sigstore/KeylessTest.java
```diff
@@ -90,7 +90,8 @@ public void sign_staging(boolean enableRekorV2) throws Exception {
     }
   }
 
-  private void verifySigningResult(List<Bundle> results) throws IOException {
+  private void verifySigningResult(List<Bundle> results, boolean expectTimestamp)
```
loosebazooka, line 93: I wonder if we can just inspect the entry type? Checking timestamp feels like something might change under us and we wont catch it.

aaronlew02: I've updated the test to check the API version instead.

loosebazooka (approved):

PR #1038, opened 2025-07-30
Use HTTP server for conformance testing

aaronlew02:
#### Summary

This change speeds up conformance testing by using an HTTP server.

### DIFF
--- a/sigstore-cli/build.gradle.kts
+++ b/sigstore-cli/build.gradle.kts
@@ -1,6 +1,9 @@
+import com.github.jengelman.gradle.plugins.shadow.tasks.ShadowJar
+
 plugins {
     id("build-logic.java")
     id("application")
+    id("com.gradleup.shadow") version "9.0.0-rc3"
 }
 
 repositories {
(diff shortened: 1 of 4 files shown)


aaronlew02:
@loosebazooka Great point. To prevent `sigstore-cli-server` distributions from being generated by `:sigstore-cli:build`, I have disabled the `shadowDistTar` and `shadowDistZip` build tasks.

sigstore-cli/build.gradle.kts
```diff
@@ -1,6 +1,9 @@
+import com.github.jengelman.gradle.plugins.shadow.tasks.ShadowJar
+
 plugins {
     id("build-logic.java")
     id("application")
+    id("com.github.johnrengelman.shadow") version "8.1.1"
```
loosebazooka, line 6: https://github.com/GradleUp/shadow use this instead. 

I would try to look up if a repo has moved before adding a plugin/dependency. This has hit me in the past more than I would've liked.

sigstore-cli/build.gradle.kts
```diff
@@ -25,6 +31,31 @@ tasks.compileJava {
...
+}
+
 tasks.run.configure {
     workingDir = rootProject.projectDir
 }
+
+tasks.register<ShadowJar>("serverShadowJar") {
+    archiveBaseName.set("sigstore-cli-server")
+    archiveClassifier.set("all")
+
```
loosebazooka, line 50: set archiveVersion to `""` so we don't have to wildcard in conformance.yaml:49 (I realize we were doing it before) but be nice to fix that now.

sigstore-cli/sigstore-cli-server
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
```
loosebazooka, line 1: can you run [shell-check](https://www.shellcheck.net/) over this?

loosebazooka (approved):
lgtm, just some minor things

loosebazooka (approved):

PR #1042, opened 2025-08-06
Configure maven central publishing

loosebazooka:
should consume secrets from gcp during publishing. Note that this means we have moved our pgp key to something else and we *must* document that releases 2.x onwards use the new key.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -1,5 +1,6 @@
 plugins {
     id("build-logic.root-build")
+    id("com.gradleup.nmcp.aggregation") version "1.0.2"
     // The Kotlin Gradle plugin was loaded multiple times in different subprojects, which is not supported and may break the build.
     `embedded-kotlin` apply false
 }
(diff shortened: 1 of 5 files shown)


build.gradle.kts
```diff
@@ -15,3 +16,17 @@ val parameters by tasks.registering {
...
+        username = providers.environmentVariable("CENTRAL_PORTAL_USERNAME")
+        password = providers.environmentVariable("CENTRAL_PORTAL_PASSWORD")
+        publishingType = "USER_MANAGED"
+        publicationName = "sigstore protobuf-specs $version"
+    }
+}
+
+dependencies {
+    nmcpAggregation(project(":sigstore-java"))
+    nmcpAggregation(project(":sigstore-maven-plugin"))
```
vlsi, line 31: Please add gradle plugin artifacts as well

vlsi (changes requested):

build.gradle.kts
loosebazooka: this is creating some issue where it's trying to apply the plugin on itself? We can try to add it in after, but for now, I'd like to get this in so we can publish these (and still publish the other to the portal)

loosebazooka: It's not exactly set up like that. `publish-to-central` only has the config to publish to central; while signing, etc is all configured in `java-published-library`

We've run into this problem before where the plugin cannot be applied on itself. I'm happy to explore this again though, currently it's blocking with an unclear solution.

loosebazooka: any update here? I would like to unblock this.

vlsi (approved):

PR #1044, opened 2025-08-07
chore: use Java 21 for building sigstore-java

vlsi:
#### Summary

This change bumps JDK we use when building the project, and it is mainly to avoid known bugs in the toolchain.

The resulting binaries should still be compatible with Java 11.

At the same time, the commit skips "./gradlew test" checks for Java 11, and we still execute tests with Java 11 via conformance.yml.

It would allow us to bump testing libraries as some of them require Java 17.

#### Release Note

* Require Java 21 when building sigstore-java. The resulting artifacts are still compatible with Java 11

#### Documentation

NONE


vlsi:
```
> Task :sigstore-java:forbiddenApisMain
Forbidden method invocation: java.net.URL#<init>(java.lang.String) [Deprecated in Java 21]
  in dev.sigstore.tuf.SigstoreTufClient$Builder (SigstoreTufClient.java:77)
Forbidden method invocation: java.net.URL#<init>(java.lang.String) [Deprecated in Java 21]
  in dev.sigstore.tuf.SigstoreTufClient$Builder (SigstoreTufClient.java:92)
Forbidden method invocation: java.net.URL#<init>(java.lang.String) [Deprecated in Java 21]
  in dev.sigstore.tuf.SigstoreTufClient$Builder (SigstoreTufClient.java:132)
Forbidden method invocation: java.net.URL#<init>(java.lang.String) [Deprecated in Java 21]
  in dev.sigstore.tuf.SigstoreTufClient$Builder (SigstoreTufClient.java:133)
Forbidden method invocation: java.net.URL#<init>(java.lang.String) [Deprecated in Java 21]
```

Should we prefer using `URI` in sigstore-java public APIs like `public Builder tufMirror(URL mirror, RootProvider trustedRoot) {`?

---

`new URI` has a new checked exception `URISyntaxException` which we probably need to wrap with the existing `MalformedURLException`

loosebazooka:
There's some confusion in sigstore-testkit/src/main/kotlin/dev/sigstore/testkit/BaseGradleTest.kt, I'm not super sure how to resolve testing compatibility.

vlsi:
Looks like ossfuzz hard-codes Java 17 in its builder image: https://github.com/google/oss-fuzz/tree/ce8bd4360b38fd6a4f83c27de5ecee2d0fe24e49/infra/base-images/base-builder-jvm

It sounds like an unfortunate decision.

vlsi:
@loosebazooka , I think this is workable except `build (11)` is still in "required CI checks".
WDYT of replacing "build (11)" required check with "build (21)"?

PR #1045, opened 2025-08-08
WIP: Use HTTP server for TUF conformance testing

aaronlew02:
#### Summary

This change speeds up TUF conformance testing by using an HTTP server.


tuf-cli/tuf-cli-server
```diff
@@ -0,0 +1,30 @@
...
+for arg in "$@"; do
+  escaped_arg=$(echo -n "$arg" | jq -R -s '.')
+  ARGS_JSON="$ARGS_JSON$escaped_arg,"
+done
+if [[ $ARGS_JSON == *, ]]; then
+  ARGS_JSON="${ARGS_JSON%,}"
+fi
+ARGS_JSON="$ARGS_JSON]"
+
+FAKETIME_VAL="${FAKETIME:-}"
```
jku, line 17: I think you could just call "date" here with useful flags, maybe `date +%s` -- or whatever works for your parsing in the java code

I suppose this would set the faketime value in your json payload on every call even if the test suite isn't setting it -- that might be ok?

jku:
btw, is it possible to increase the verbosity of the client code somehow? Doesn't have to be a CLI flag or anything if that's complicated, a code change is fine,  but would be useful to be able to see a little more of the client decisions...

tuf-cli/tuf-cli-server
aaronlew02: This works, thank you!

tuf-cli/src/main/java/dev/sigstore/tuf/cli/TestClock.java
```diff
@@ -0,0 +1,34 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package dev.sigstore.tuf.cli;
+
+import java.time.Clock;
+
+public class TestClock {
```
loosebazooka, line 20: I'm not super fond of the idea of a global test clock. I think this should be instantiable . And we can modify the cli to accept a "time" with a --time flag.

loosebazooka:
lets merge this for now, but I'd like to find a way to only keep one xfails file. So lets open an issue or update the build immediate after this.

tuf-cli/build.gradle.kts
```diff
@@ -15,6 +18,11 @@ dependencies {
     implementation(platform("com.google.oauth-client:google-oauth-client-bom:1.39.0"))
     implementation("com.google.oauth-client:google-oauth-client")
 
+    implementation("org.eclipse.jetty:jetty-server:11.0.24")
+    implementation("org.eclipse.jetty:jetty-servlet:11.0.24")
+
+    implementation("org.slf4j:slf4j-simple:2.0.17")
```
loosebazooka, line 24: was this necessary?

loosebazooka: I don't see any logging code anywhere?

loosebazooka (approved):

tuf-cli/build.gradle.kts
aaronlew02: This is used for the regular conformance server as well.

Starting the server without the implementation produces the following:
```
SLF4J: No SLF4J providers were found.
SLF4J: Defaulting to no-operation (NOP) logger implementation
SLF4J: See https://www.slf4j.org/codes.html#noProviders for further details.
```

Starting the server with the implementation produces the following:
```
[main] INFO org.eclipse.jetty.server.Server - jetty-11.0.24; built: 2024-08-26T18:11:22.448Z; git: 5dfc59a691b748796f922208956bd1f2794bcd16; jvm 21.0.8+12-LTS-jvmci-23.1-b72
[main] INFO org.eclipse.jetty.server.AbstractConnector - Started ServerConnector@3a82f6ef{HTTP/1.1, (http/1.1)}{0.0.0.0:8080}
[main] INFO org.eclipse.jetty.server.Server - Started Server@146ba0ac{STARTING}[11.0.24,sto=0] @153ms
```

PR #1061, opened 2025-09-01
tuf Updater: fix snapshot version rollback case

jku:
The snapshot version listed in timestamp must never decrease (except if timestamp keys are rotated but that in that case local timestamp is not used at all).

This was a deviation from TUF specification but not a very serious one in context: In sigstores TUF repository the Timestamp/snapshot keys are exposed in the same environments so compromise in one means compromise for both (because of this they are in fact the same key). So tricks like snapshot version rollback are not that relevant.


jku:
The logs are a little light on details but my first guess is this is something unrelated?

```
2025-09-01T05:53:00.2039522Z Caused by: dev.sigstore.oidc.client.OidcException: Could not find an oidc provider
2025-09-01T05:53:00.2040101Z 	at dev.sigstore.oidc.client.OidcClients.getIDToken(OidcClients.java:65)
2025-09-01T05:53:00.2040656Z 	at dev.sigstore.KeylessSigner.renewSigningCertificate(KeylessSigner.java:593)
2025-09-01T05:53:00.2041187Z 	at dev.sigstore.KeylessSigner.sign(KeylessSigner.java:420)
2025-09-01T05:53:00.2041539Z 	... 27 more
```

jku:
oh, I guess the example test uses the workflow identity to sign so they just fail for all PRs from forks

loosebazooka:
> oh, I guess the example test uses the workflow identity to sign so they just fail for all PRs from forks

Yeah, that's something I've been meaning to fix, but not pressing.

PR #1063, opened 2025-09-04
fix: Reject unsupported DSSE version

aaronlew02:
#### Summary

This change adds an exception that is thrown when the version of a DSSE envelope in a Rekor entry is unsupported.


loosebazooka (dismissed):

loosebazooka:
this PR seems to have a weird commit state actually, might be worth doing a clean rebase

PR #1076, opened 2025-09-24
remove oidc config from gradle plugin

loosebazooka:
- non-ambient credentials should be configured by signing-config
- ambient credentially are now unmodifiable defaults (behavior that is consistent with other clients)


loosebazooka:
@vlsi if no objections, I'll probably just merge this for v2 release of sigstore-java (and plugins)

aaronlew02 (approved):

vlsi (approved):

PR #1107, opened 2025-11-11
test: improve Gradle compatibility matrix

vlsi:
Use Java home from testJdkVersion to launch test Gradle instances

See https://github.com/sigstore/sigstore-java/pull/1105#issuecomment-3517915305


loosebazooka:
gated by id-token:write ?

PR #1109, opened 2025-11-14
workflows: Run conformance suites in parallel

### LINKED ISSUE #1110: run conformance suites in parallel
opened a PR for tuf-conformance already: https://github.com/sigstore/sigstore-java/pull/1109

The sigstore-conformance workflow still has "max-parallel: 1" because at some point staging had infra issues that lead to high latency/timeouts if multiple workflows were running at same time: would make sense to test full parallelism again and maybe start debugging a bit further if those issues persist

jku:
Closes #1110

This should be just fine?

For reference, the sigstore-conformance workflow has "max-parallel: 1" because at some point staging had infra issues that lead to high latency/timeouts if multiple workflows were running at same time -- tuf-conformance should not have issues like that as far as I can tell.

### DIFF
--- a/.github/workflows/conformance.yml
+++ b/.github/workflows/conformance.yml
@@ -12,7 +12,6 @@ on:
 jobs:
   conformance:
     strategy:
-      max-parallel: 1
       matrix:
         java-version: [11, 17, 21]
         sigstore-env: [production, staging]


loosebazooka:
examples failing is from a fork
that's fine.

aaronlew02:
Thank you, @jku!

The TUF conformance tests appear to run fine in parallel, so I parallelized the regular conformances tests here as well.

PR #1126, opened 2025-12-18
workflows: schedule a weekly tuf-conformance run

jku:
This makes sure there's an up-to-date result
for https://theupdateframework.github.io/tuf-conformance/

There could also be automation to file an issue on failure but that's not included in this change (and I think failures like that should be very unlikely for tuf-conformance)

### DIFF
--- a/.github/workflows/tuf-conformance.yml
+++ b/.github/workflows/tuf-conformance.yml
@@ -7,7 +7,8 @@ on:
     branches: [main]
   pull_request: {}
   workflow_dispatch: {}
-  # TODO: add cron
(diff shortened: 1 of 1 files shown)


.github/workflows/tuf-conformance.yml
```diff
@@ -7,7 +7,8 @@ on:
     branches: [main]
   pull_request: {}
   workflow_dispatch: {}
-  # TODO: add cron
+  schedule:
```
loosebazooka: Hrm.. maybe I should make this PR so you don't get pinged?

PR #1148, opened 2026-01-08
add nonce parameter to OIDC flow

bobcallaway:
# Summary:
This change hardens the OIDC authentication process by implementing a cryptographically secure nonce (number used once) in the WebOidcClient.

# Key Changes:
   * Nonce Generation: Introduces a generateNonce() method that creates a random, URL-safe string using SecureRandom.
   * Request Binding: Updates the authorization flow to include this nonce in the initial authorization request URL.
   * Validation: Adds strict validation logic to verify that the nonce claim in the returned ID token matches the original nonce sent by the client.
   * Testing: Adds WebOidcClientNonceTest to verify that tokens with correct nonces are accepted and tokens with mismatched or missing nonces are rejected.

# Security Benefit:
This mechanism is a critical defense against Replay Attacks. By binding the ID token to the specific client request that initiated it, we ensure that an attacker cannot capture a valid ID token from a previous session and "replay" it to impersonate a user. If the nonces do not match, the token is immediately rejected.

### DIFF
--- a/sigstore-java/src/test/java/dev/sigstore/oidc/client/WebOidcClientNonceTest.java
+++ b/sigstore-java/src/test/java/dev/sigstore/oidc/client/WebOidcClientNonceTest.java
@@ -18,15 +18,14 @@ package dev.sigstore.oidc.client;
 import com.gargoylesoftware.htmlunit.WebClient;
 import com.google.common.io.Resources;
 import dev.sigstore.trustroot.Service;
+import java.io.IOException;
+import java.nio.charset.StandardCharsets;
+import java.util.Map;
 import no.nav.security.mock.oauth2.MockOAuth2Server;
 import no.nav.security.mock.oauth2.OAuth2Config;
 import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.Test;
-
-import java.io.IOException;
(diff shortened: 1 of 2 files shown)


sigstore-java/src/test/java/dev/sigstore/oidc/client/WebOidcClientNonceTest.java
```diff
@@ -0,0 +1,83 @@
...
+
+import java.io.IOException;
+import java.nio.charset.StandardCharsets;
+import java.util.Map;
+
+public class WebOidcClientNonceTest {
+
+  private MockOAuth2Server server;
+
+  @AfterEach
```
loosebazooka: this is automatically handled by the server extension (MockOAuth2ServerExtension)

loosebazooka (commented):
looks fine, gonna run some tests locally just to double check the web flow still works.

jku (commented):
I think this should work -- `nonce` is optional but should always be returned in the response if it is included in the request.

That said `state` is marked as recommended for the authorization code flow and seems to be for a similar purpose so this could use state instead. The result seems about the same but maybe that requires less work since AuthorizationCodeRequestUrl has a `setState()`  constructor chain method (I can't quite see how that all works so could be wrong and it's not easier at all)

PR #1173, opened 2026-02-23
Delegated targets support

loosebazooka:
- No support for succinct delegations
- Removed eager download of all targets
  - SigstoreTufClient must now download the targets it wants explicitly
- Rename update() to refresh() to match naming of other clients
- Correctly handle url escaping in names

NOTE: this was produced with the help of AI coding tools


sigstore-java/src/main/java/dev/sigstore/tuf/Updater.java
```diff
@@ -562,6 +532,254 @@ void downloadTarget(String targetName, TargetData targetData) throws IOException
...
+
+    Optional<Targets> localTargets = trustedMetaStore.findTargets(roleName);
+    if (localTargets.isPresent()) {
+      if (localTargets.get().getSignedMeta().getVersion() == snapshotTarget.getVersion()) {
+        return localTargets.get();
+      }
+      if (localTargets.get().getSignedMeta().getVersion() > snapshotTarget.getVersion()) {
+        throw new SnapshotTargetVersionException(
+            roleName, snapshotTarget.getVersion(), localTargets.get().getSignedMeta().getVersion());
+      }
```
jku, line 764: two possible issues here:
* if a local targets file is present... it should still be verified (hashes match, threshold of signatures is correct), right?
  * If these checks are not done for other local file reads either, then no need to make this specific one more complicated in this PR
  * If you're wondering why verify local "trusted" files: It potentially makes the client a bit more secure (even if local attack is tricky to defend against) but it also makes client safer against cache corruption and the rare case mentioned below (where local targets may no longer be signed by the correct keys and version numbers may have restarted)
  * The conformance test suite does not try to test local cache corruption issues since it would require fully enforcing a specific storage format which feels wrong -- if someone wants to use sqlite they should be able to...
* targets version could in a very rare case go down (see the note in 5.3.11 in TUF spec: if the "snapshot memory" is wiped, repository is free to e.g. restart targets versions from 1 -- this seems like a case your code supports elsewhere) so the second if-clause seems unnecessary

jku (approved):
Looks good to me:
* I would have raised the question of _"do you really want more code in sigstore-java"_ but if there are any plans of extracting the tuf part into a library, then I think this is valid as is it makes the implementation spec complete
* I've read the code, left one note about reading local metadata 
* I would not trust my ability to verify this code by just reading (since I've overlooked so many bugs in the corresponding python implementation) but I am much more confident about the conformance testing: it really should cover a lot of potential mistakes in this area

sigstore-java/src/main/java/dev/sigstore/tuf/Updater.java
jku: the comment looks a bit dumb, I think I failed to select correct context: it's supposed to  reference **both** if-clauses at the top of the method

loosebazooka: right, I had this fixed. Should be okay for now.

PR #1176, opened 2026-03-04
http fulcio client

loosebazooka:
#### Summary
Seem like private deployments don't want to expose the grpc endpoint, so provide the http client.


sigstore-java/src/main/java/dev/sigstore/fulcio/client/FulcioClientHttp.java
```diff
@@ -0,0 +1,184 @@
...
+                Locale.ROOT,
+                "bad response from fulcio @ '%s' : %s",
+                endpoint,
+                resp.parseAsString()));
+      }
+      responseJson = resp.parseAsString();
+    } catch (IOException e) {
+      throw new CertificateException("Failed to request signing certificate from fulcio", e);
+    }
+
```
aaronlew02, line 139: `execute()` will return an `HttpResponseException` "when an error status code is detected in an HTTP response" ([docs](https://docs.cloud.google.com/java/docs/reference/google-http-client/latest/com.google.api.client.http#httpresponseexception)). We should either catch the `HttpResponseException` like [here](https://github.com/sigstore/sigstore-java/blob/a52ffe1fbce255058546c7c2848c30b292028b4e/sigstore-java/src/main/java/dev/sigstore/rekor/client/RekorClientHttp.java#L123) or disable the exception like [here](https://github.com/sigstore/sigstore-java/blob/a52ffe1fbce255058546c7c2848c30b292028b4e/sigstore-java/src/main/java/dev/sigstore/tuf/HttpFetcher.java#L58) and [here](https://github.com/sigstore/sigstore-java/blob/a52ffe1fbce255058546c7c2848c30b292028b4e/sigstore-java/src/main/java/dev/sigstore/timestamp/client/TimestampClientHttp.java#L120).

aaronlew02:
This looks great, though I'm a bit concerned about duplication between the gRPC and HTTP client implementations as well as their tests. For the clients, WDYT about extracting common logic into a utility class or moving it into `FulcioClient`? For the tests, WDYT about parameterizing the original tests instead of creating new ones?

loosebazooka:
adjusted testing and moved cert parsing.

aaronlew02 (approved):

PR #1182, opened 2026-04-15
Replace URI.resolve with URIFormat.appendPath

### LINKED ISSUE #1005: check uses of URI.resolve
URI resolve works right now in sigstore-java because of how we distribute our base urls, but that could change. We should either write a custom handler or look deeper into how to fortify this a bit.

aaronlew02:
Closes #1005

#### Summary

This change replaces all calls to `URI.resolve` with a call to the custom URI handler `URIFormat.appendPath`, to predictably handle trailing slashes in base URLs.


loosebazooka:
we should also ban the use of URI.resolve in the codebase and force devs to use appendPath. check `config/forbiddenApis.txt` and there's a way to apply exceptions.

loosebazooka (approved):

PR #1185, opened 2026-05-04
Re-add and enhance SET verification in KeylessVerifier

aaronlew02:
#### Summary

This change re-adds SET verification in `KeylessVerifier`, which was erroneously removed in #1008, and enhances it for spec compliance such that the integration time is:
* trusted only if an inclusion promise is present ([source](https://github.com/sigstore/protobuf-specs/blob/d4c77e3f4a5d73802cf3e4d807379c27385d5488/protos/sigstore_rekor.proto#L104-L105))
* required to be verified if no other form of time is present ([source](https://github.com/sigstore/protobuf-specs/blob/d4c77e3f4a5d73802cf3e4d807379c27385d5488/protos/sigstore_rekor.proto#L112)) 
* verified along with any other sources of time present ([source](https://github.com/sigstore/protobuf-specs/blob/d4c77e3f4a5d73802cf3e4d807379c27385d5488/protos/sigstore_rekor.proto#L113))


sigstore-java/src/main/java/dev/sigstore/KeylessVerifier.java
```diff
@@ -193,9 +193,17 @@ public void verify(byte[] artifactDigest, Bundle bundle, VerificationOptions opt
   }
 
   private void verifyTimestamps(
-      X509Certificate leafCert, List<Bundle.Timestamp> timestamps, byte[] signature)
+      X509Certificate leafCert, List<Bundle.Timestamp> timestamps, java.time.Instant entryTime, byte[] signature)
       throws KeylessVerificationException {
     if (timestamps == null || timestamps.isEmpty()) {
+      var entryDate = java.util.Date.from(entryTime);
```
loosebazooka: is entryTime guaranteed to be non-null?

aaronlew02: Yes, because `getIntegratedTimeInstant()` always returns a `long`. However, that some form of verifiable time (SET or RFC 3161 timestamp) is provided is already [checked](https://github.com/sigstore/sigstore-java/blob/6d736e2f49fce2ae2b3127471acb475639a69912/sigstore-java/src/main/java/dev/sigstore/bundle/Bundle.java#L81) in `Bundle.java`.

loosebazooka: The error message could be confusing if it's a v2 style log entry and entryTime is 0?

loosebazooka: Also I think we're required to verify all times if they are valid. Especially for v1, even if they provide an rfc3161 timestamp?

aaronlew02: I've updated the logic so that:
* in `checkAtLeastOneTimestamp()`, only a SET with an entry time >0 is considered verifiable
* in `verifyTimestamps()`, all verifiable time is verified

I had to update `testVerify_invalidSet_validRfc3161Timestamp_rekorV1()` so that it will continue to target the invalid SET case instead of the case where the SET predates the certificate validity.

sigstore-java/src/main/java/dev/sigstore/KeylessVerifier.java
```diff
@@ -183,7 +183,8 @@ public void verify(byte[] artifactDigest, Bundle bundle, VerificationOptions opt
       signature = dsseEnvelope.getSignature();
     }
 
-    verifyTimestamps(leafCert, bundle.getTimestamps(), signature);
```
loosebazooka, line 186: I know there are checks in bundle object creation to enforce the existence of one timestamp, however, if that were to change under us we wouldn't know. I think we should check somewhere in here that there is at least one timestamp and then validate them all.

sigstore-java/src/main/java/dev/sigstore/KeylessVerifier.java
```diff
@@ -193,10 +194,20 @@ public void verify(byte[] artifactDigest, Bundle bundle, VerificationOptions opt
   }
 
   private void verifyTimestamps(
-      X509Certificate leafCert, List<Bundle.Timestamp> timestamps, byte[] signature)
+      X509Certificate leafCert,
+      List<Bundle.Timestamp> timestamps,
+      java.time.Instant entryTime,
```
loosebazooka: Fully qualified usage of java.time.xyz seems odd to me.

sigstore-java/src/main/java/dev/sigstore/KeylessVerifier.java
```diff
@@ -193,10 +194,20 @@ public void verify(byte[] artifactDigest, Bundle bundle, VerificationOptions opt
...
+      byte[] signature)
       throws KeylessVerificationException {
-    if (timestamps == null || timestamps.isEmpty()) {
-      return;
+    if (!entryTime.equals(java.time.Instant.EPOCH)) {
```
loosebazooka: There is somewhere in the spec that states that we only inspect entryTime if the SET is valid. So maybe we can use that somehow to manage this condition. Rather than comparing to 0 or whatever?

sigstore-java/src/main/java/dev/sigstore/KeylessVerifier.java
aaronlew02: I've added validation to `verifyTimestamps` that at least one valid timestamp (of any form) is found in the bundle.

sigstore-java/src/main/java/dev/sigstore/KeylessVerifier.java
aaronlew02: I've chronologically switched timestamp verification against the artifact with Rekor verification in the workflow. That way, we know that if a SET is present, it has been validated to be verified against the artifact according to the spec. The condition in question then becomes one of whether the entry time from a validated SET has been provided.

sigstore-java/src/main/java/dev/sigstore/KeylessVerifier.java
```diff
@@ -183,20 +184,35 @@ public void verify(byte[] artifactDigest, Bundle bundle, VerificationOptions opt
...
     }
 
-    verifyTimestamps(leafCert, bundle.getTimestamps(), signature);
-
     try {
       rekorVerifier.verifyEntry(rekorEntry);
     } catch (RekorVerificationException ex) {
       throw new KeylessVerificationException("Transparency log entry could not be verified", ex);
     }
+
```
loosebazooka, line 193: I think a comment here could be helpful for someone in the future reading this code.

```
// if entry was verified and has a set, get time from it
```

sigstore-java/src/main/java/dev/sigstore/KeylessVerifier.java
```diff
@@ -183,20 +184,35 @@ public void verify(byte[] artifactDigest, Bundle bundle, VerificationOptions opt
...
+    if (entryTime != null) {
+      var entryDate = java.util.Date.from(entryTime);
+      try {
+        leafCert.checkValidity(entryDate);
+      } catch (CertificateNotYetValidException e) {
+        throw new KeylessVerificationException("Signing time was before certificate validity", e);
+      } catch (CertificateExpiredException e) {
+        throw new KeylessVerificationException("Signing time was after certificate expiry", e);
+      }
+    } else if (timestamps.isEmpty()) {
```
loosebazooka: For readability: this `else if` could be a guard clause up top before checking any of the timestamps.

```
if (timestamps.isEmpty() && entryTime == null) {
 throw new KeylessVerificationException("No timestamps found in bundle");
}
```

sigstore-java/src/test/java/dev/sigstore/KeylessVerifierTest.java
```diff
@@ -774,7 +774,7 @@ public void testVerify_invalidSet_validRfc3161Timestamp_rekorV1() throws Excepti
 
     var invalidBundleFile =
         bundleFileWithTs.replace(
-            "\"integratedTime\": \"1747928754\"", "\"integratedTime\": \"1747928000\"");
```
loosebazooka, line 777: does this need to change?

loosebazooka (commented):
looks good, just some minor things

sigstore-java/src/test/java/dev/sigstore/KeylessVerifierTest.java
aaronlew02: Good catch. Not anymore, since Rekor and SET verification have been switched chronologically. I've undone this change.

sigstore-java/src/main/java/dev/sigstore/KeylessVerifier.java
```diff
@@ -183,20 +184,37 @@ public void verify(byte[] artifactDigest, Bundle bundle, VerificationOptions opt
...
+      Instant entryTime,
+      byte[] signature)
       throws KeylessVerificationException {
-    if (timestamps == null || timestamps.isEmpty()) {
-      return;
+    if (timestamps.isEmpty() && entryTime == null) {
+      throw new KeylessVerificationException("No valid timestamps found in bundle");
+    }
+    if (entryTime != null) {
+      var entryDate = java.util.Date.from(entryTime);
```
loosebazooka: Date could probably be an import too

loosebazooka (commented):
Lgtm. This should have a test though especially to prevent regression.

loosebazooka (approved):

PR #1197, opened 2026-05-27
Make algorithm registry more central to system

loosebazooka:
This is more setup for using any hash algorithm in the system. But currently still limit to sha256.

### DIFF
--- a/sigstore-java/src/main/java/dev/sigstore/AlgorithmRegistry.java
+++ b/sigstore-java/src/main/java/dev/sigstore/AlgorithmRegistry.java
@@ -15,11 +15,55 @@
  */
 package dev.sigstore;
 
-import static com.google.common.hash.Hashing.*;
+import static com.google.common.hash.Hashing.sha256;
 
 import com.google.common.hash.HashFunction;
+import java.security.PublicKey;
+import java.security.interfaces.ECPublicKey;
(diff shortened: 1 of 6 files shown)


sigstore-java/src/main/java/dev/sigstore/AlgorithmRegistry.java
```diff
@@ -15,11 +15,54 @@
...
+    if ("EC".equals(algorithm) || "ECDSA".equals(algorithm)) {
+      if (publicKey instanceof ECPublicKey) {
+        var ecKey = (ECPublicKey) publicKey;
+        int fieldSize = ecKey.getParams().getCurve().getField().getFieldSize();
+        if (fieldSize == 256) {
+          return SigningAlgorithm.PKIX_ECDSA_P256_SHA_256;
+        }
+        throw new UnsupportedAlgorithmException("Unsupported EC field size: " + fieldSize);
+      }
+      throw new UnsupportedAlgorithmException("EC key must be an instance of ECPublicKey");
```
aaronlew02: Since this depends on a deeper issue converting `publicKey` to the `ECPublicKey`, we should use a different exception type like `RuntimeException`.

aaronlew02 (changes requested):
LGTM. One small comment.

aaronlew02 (approved):

PR #1198, opened 2026-06-01
Support more hash/signing algorithms

loosebazooka:


loosebazooka:
actually, the signing algorithm mapping only really applies to rekor v2 dsseXhashedrekord. We have some legacy bundles that mix and match keys + hash algorithm. I need to rethink this whole thing

loosebazooka:
okay better, this is ready for review

sigstore-java/src/main/java/dev/sigstore/encryption/signers/RsaVerifier.java
aaronlew02, line 59: Having a configurable `hashAlgorithm` in this class, while also forcing `PKSC1_SHA256_PADDING` and only ever [instantiating](https://github.com/sigstore/sigstore-java/blob/2adbbf886ec36673bd5e0cfe83fcf1ec2ccc0c92/sigstore-java/src/main/java/dev/sigstore/encryption/signers/Verifiers.java#L28-L30) it with `SHA2_256`, might send mixed signals.


Can we enforce `this.HashAlgorithm = AlgorithmRegistry.HashAlgorithm.SHA2_256` [above](https://github.com/sigstore/sigstore-java/blob/2adbbf886ec36673bd5e0cfe83fcf1ec2ccc0c92/sigstore-java/src/main/java/dev/sigstore/encryption/signers/RsaVerifier.java#L29) and only allow single-argument construction of `RsaVerifier` with `publicKey`?

sigstore-java/src/main/java/dev/sigstore/encryption/Hashers.java
```diff
@@ -0,0 +1,37 @@
...
+
+public class Hashers {
+  public static HashFunction from(AlgorithmRegistry.HashAlgorithm hashAlgorithm) {
+    switch (hashAlgorithm) {
+      case SHA2_256:
+        return com.google.common.hash.Hashing.sha256();
+      case SHA2_384:
+        return com.google.common.hash.Hashing.sha384();
+      case SHA2_512:
+        return com.google.common.hash.Hashing.sha512();
```
aaronlew02: Can we import `com.google.common.hash.Hashing` instead of using qualified references?

aaronlew02 (changes requested):
This looks great! Just a couple of comments.

loosebazooka:
okay switched rsa to sha256 only

aaronlew02 (approved):

PR #1221, opened 2026-06-15
Spotless update

loosebazooka:
Not sure if this is the most elegant solution, but seems to be what is required for this problem. We might be able to share version information via a toml version catalog. But hopefully updates happen as a group

### DIFF
--- a/build-logic/jvm/build.gradle.kts
+++ b/build-logic/jvm/build.gradle.kts
@@ -9,7 +9,7 @@ repositories {
 dependencies {
     implementation(project(":basics"))
     implementation(project(":build-parameters"))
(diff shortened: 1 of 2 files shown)


vlsi:
I suggest automatic grouping like in https://github.com/pgjdbc/pgjdbc/pull/4184 and https://github.com/apache/jmeter/pull/6710

loosebazooka:
@vlsi does the automatic grouping pattern matching for renovate handle plugin matching?

vlsi:
Renovate should group them automatically, and it does group when I execute renovate locally.

loosebazooka:
okay rebased, should be ready for review

vlsi (approved):

PR #1242, opened 2026-07-07
Move to apache-v5 in google http client

loosebazooka:

### DIFF
--- a/sigstore-java/src/main/java/dev/sigstore/http/HttpClients.java
+++ b/sigstore-java/src/main/java/dev/sigstore/http/HttpClients.java
@@ -18,14 +18,17 @@ package dev.sigstore.http;
 import com.google.api.client.http.HttpBackOffIOExceptionHandler;
 import com.google.api.client.http.HttpRequestFactory;
 import com.google.api.client.http.HttpTransport;
-import com.google.api.client.http.apache.v2.ApacheHttpTransport;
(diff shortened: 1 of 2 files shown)


sigstore-java/src/main/java/dev/sigstore/http/HttpClients.java
```diff
@@ -36,11 +39,24 @@ public class HttpClients {
...
    */
   public static HttpTransport newHttpTransport(HttpParams httpParams) {
-    HttpClientBuilder hcb = ApacheHttpTransport.newDefaultHttpClientBuilder();
     if (httpParams.getAllowInsecureConnections()) {
-      hcb.setSSLHostnameVerifier(NoopHostnameVerifier.INSTANCE);
+      var connManager =
+          PoolingHttpClientConnectionManagerBuilder.create()
+              .setSSLSocketFactory(
+                  SSLConnectionSocketFactoryBuilder.create()
+                      .setHostnameVerifier(NoopHostnameVerifier.INSTANCE)
```
vlsi, line 47: Just noticed. Did we really intend a security issue here? Should we verify hostname for the SSL?

loosebazooka: this is only for the insecure option. like if you want to hit a local server with no tls or something.

vlsi (approved):

aaronlew02 (approved):

PR #1244, opened 2026-07-17
Add java25 checks and unit test min to java17

loosebazooka:
- Set testCompile tasks to target java17
- Also compile sigstore-testkit module with java17
- Upgrade mock-oauth2-server to 5.0.2
- Update google java format to 1.35.0

### DIFF
--- a/build-logic/jvm/src/main/kotlin/build-logic.java.gradle.kts
+++ b/build-logic/jvm/src/main/kotlin/build-logic.java.gradle.kts
@@ -25,7 +25,7 @@ tasks.configureEach<JavaExec> {
 
 spotless {
     java {
-        googleJavaFormat("1.24.0")
+        googleJavaFormat("1.35.0")
         licenseHeaderFile("$rootDir/config/licenseHeader")
         // Note if submodule needs to add more exclusions, it should list ALL of them since
         // Spotless does not have "addTargetExclude" method
--- a/build-logic/jvm/src/main/kotlin/build-logic.java.gradle.kts
+++ b/build-logic/jvm/src/main/kotlin/build-logic.java.gradle.kts
@@ -25,7 +25,7 @@ tasks.configureEach<JavaExec> {
 
 spotless {
     java {
-        googleJavaFormat("1.24.0")
+        googleJavaFormat("1.35.0")
         licenseHeaderFile("$rootDir/config/licenseHeader")
         // Note if submodule needs to add more exclusions, it should list ALL of them since
         // Spotless does not have "addTargetExclude" method
@@ -36,14 +36,17 @@ spotless {
 tasks.withType<JavaCompile>().configureEach {
     inputs.property("java.version", System.getProperty("java.version"))
     inputs.property("java.vm.version", System.getProperty("java.vm.version"))
-    options.apply {
-        encoding = "UTF-8"
-        compilerArgs.add("-Xlint:deprecation")
-        if (buildParameters.failOnJavacWarning) {
-            compilerArgs.add("-Werror")
-        }
+    options.encoding = "UTF-8"
+    options.compilerArgs.add("-Xlint:deprecation")
+    if (buildParameters.failOnJavacWarning) {
+        options.compilerArgs.add("-Werror")
+    }
+    options.release.set(buildParameters.targetJavaVersion)
+}
 
-        release.set(buildParameters.targetJavaVersion)
+plugins.withId("java") {
+    tasks.named<JavaCompile>("compileTestJava") {
+        options.release.set(17)
     }
 }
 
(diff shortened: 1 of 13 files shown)


sigstore-java/src/test/java/dev/sigstore/oidc/client/WebOidcClientTest.java
```diff
@@ -23,9 +23,12 @@
 import java.util.Map;
 import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.condition.DisabledForJreRange;
+import org.junit.jupiter.api.condition.JRE;
 import org.junit.jupiter.api.extension.RegisterExtension;
 import org.slf4j.event.Level;
 
+@DisabledForJreRange(max = JRE.JAVA_16)
```
vlsi: It it the same as `@EnabledForJreRange(min = JRE.JAVA_17)`?
Frankly, I find enabled easier to read.

```suggestion
@EnabledForJreRange(min = JRE.JAVA_17)
```

sigstore-java/build.gradle.kts
```diff
@@ -10,6 +10,22 @@ plugins {
...
+        17,
+    )
+}
+configurations.testRuntimeClasspath.configure {
+    attributes.attribute(
+        org.gradle.api.attributes.Attribute
+            .of("org.gradle.jvm.version", Int::class.javaObjectType),
+        17,
+    )
+}
```
vlsi: These changes look odd. Why do we need explicit attributes?

vlsi (dismissed):

sigstore-java/build.gradle.kts
loosebazooka: There's a few options here. This just allowed the mockoidcclient (min java17) to be allowed on the classpath even if the testjdk was java11.

I think it can be cleaned up a little syntax wise. The alternatives are to isolate out those tests in other ways, or to just move our minimum to java17.

vlsi: We can't pretend test sources to be "compatible wit java 11" while Mockito requires 17, can we?
If Mockito requires Java 17, we will might get compilation or runtime exceptions any time soon..

If we use Mockito in a few tests only, we might consider moving them to a separate module and/or sourceset (so it has its own compilation unit with its own target release=17).

loosebazooka: I think this is the trick for it. We're telling gradle to just allow it, and it seems to fine doing so. If the classes were ever loaded it would fail, but they're gated by the junit annotations.

loosebazooka: I think moving to java17 as the minimum might be fine, we can remove java11 from our standard tests and just run it all through conformance testing.

vlsi: I'm afraid compilation might fail. The code should be compiled first, and the compiler needs to parse the bytecode. It shouldn't be able to recognize Java 17 bytecode if it targets Java 11.

loosebazooka: I mean it's not failing right now -- but I can rework this to just be java17. I would also like to explore updating to junit6 (in another PR)

loosebazooka: I think I may be misunderstanding what's going on though.

loosebazooka: oh right our build system is jdk21 targetting java11 bytecode, so it doesn't appear to fail on the java 17 reference.

loosebazooka: that being said, it sounds like this is confusing and weird. So I'll try to avoid it.

aaronlew02 (dismissed):

sigstore-java/build.gradle.kts
loosebazooka: since we [don't even run java 11 tests in CI](https://github.com/sigstore/sigstore-java/blob/main/.github/workflows/ci.yaml#L22), I just bumped the whole test compile to java17

build-logic/jvm/src/main/kotlin/build-logic.java.gradle.kts
```diff
@@ -36,14 +36,17 @@ spotless {
...
+    if (buildParameters.failOnJavacWarning) {
+        options.compilerArgs.add("-Werror")
+    }
+    options.release.set(buildParameters.targetJavaVersion)
+}
 
-        release.set(buildParameters.targetJavaVersion)
+plugins.withId("java") {
+    tasks.named<JavaCompile>("compileTestJava") {
+        options.release.set(17)
```
vlsi, line 49: It might make sense to create a property for testTargetJdk like we have for the main code

vlsi (approved):
Looks good. A single property instead of replicating 17 would be nice as well

build-logic/jvm/src/main/kotlin/build-logic.java.gradle.kts
loosebazooka: Gotcha. Gonna do in a followup. This is blocking some other PRs.

PR #1252, opened 2026-07-23
Close #1251: Allow no transparency-log entries

### LINKED ISSUE #1251: `KeylessVerifier` cannot verify bundles that use an RFC 3161 timestamp and no transparency-log entry
_I used help from AI to create this issue, but I humanly read it and reviewed/corrected it before submitting it_

### Why did I create this issue
The concrete case that surfaced this: **GitHub artifact attestations produced for a *private* repository**. Those are signed by GitHub's own (non–public-good) Sigstore instance, which does **not** write to a public transparency log — the bundle carries a TSA timestamp instead. `gh attestation verify` and `cosign` (with `--use-signed-timestamps --insecure-ignore-tlog`) both verify these bundles; `sigstore-java` cannot.

### Summary
`KeylessVerifier.verify(...)` unconditionally requires at least one Rekor transparency-log entry in the bundle. Sigstore bundles whose verification material contains **only** an RFC 3161 signed timestamp `timestampVerificationData`) and **no** `tlogEntries` are valid per the bundle spec, but verification fails before the timestamp or signature is ever checked.

  ### Related issues (this is not a duplicate)
  - **#699** added RFC 3161 TSA validation, but only as the *source of trusted time*; the Rekor entry is still mandatory (`KeylessVerifier` calls `bundle.getEntries().get(0)`).
  This issue is about verifying bundles that carry **no `tlogEntries` at all** — only a TSA timestamp.
  - **#980** requires at least one timestamp (SET or RFC 3161) but still assumes a tlog entry is present (Rekor v1 SET vs Rekor v2 + RFC 3161). The private-instance bundles here
  have zero tlog entries.
  - **#813** concerns *uploading* attestations to GitHub (DSSE), not verifying tlog-less bundles.
  - **#374** is error-messaging when rekor entries are absent, not support for verifying without them.

  ### Steps to reproduce
  1. Obtain a bundle with `timestampVerificationData` and no `tlogEntries` — e.g. an `actions/attest-build-provenance` attestation from a **private** GitHub repo (`gh attestation
  download …`), plus the matching trust root (`gh attestation trusted-root`).
  2. `sigstore-cli verify <artifact> --bundle <bundle> --trusted-root <root> --certificate-identity <san> --certificate-oidc-issuer https://token.actions.githubusercontent.com`

  ### Actual behavior
  Verification throws while fetching the (absent) Rekor entry — `bundle.getEntries().get(0)` on an empty list — so it never reaches signature/timestamp verification.

  ### Root cause
  In `sigstore-java/src/main/java/dev/sigstore/KeylessVerifier.java`, the Rekor entry is treated as mandatory:

  ```java
  // findHashAlgorithm(...)
  var rekorEntry = bundle.getEntries().get(0);
  ...
  // verify(...)
  var rekorEntry = bundle.getEntries().get(0);
  ...
  rekorVerifier.verifyEntry(rekorEntry);   // "Transparency log entry could not be verified"
  ```

  `verifyTimestamps(...)` already knows how to verify RFC 3161 timestamps and even guards `if (timestamps.isEmpty() && entryTime == null)`, but it is only reached *after* the
  mandatory Rekor entry has been obtained and verified — so a TSA-only bundle can never get there.

  ### Expected behavior
  When a bundle has no `tlogEntries` but does carry a valid `timestampVerificationData` (RFC 3161 timestamp from a TSA in the trusted root), `KeylessVerifier` should verify the Fulcio chain + DSSE/message signature and establish signing time from the TSA timestamp — i.e. treat the transparency-log entry as optional when signed timestamps are present, rather than requiring one. This matches the Sigstore bundle spec, whose verification material permits tlog entries **and/or** signed timestamps.

  ### Suggested fix
  - Make `findHashAlgorithm` / `verify` tolerate an empty `bundle.getEntries()`:
   - derive the hash algorithm without a Rekor entry when none is present (e.g. from the DSSE payload / message-signature algorithm), and
   - skip `rekorVerifier.verifyEntry(...)` when there is no entry, relying on `verifyTimestamps(...)` (TSA) for the trusted signing time.
  - Keep requiring **at least one** source of trusted time (tlog SET *or* RFC 3161 timestamp), which `verifyTimestamps` already enforces.
  
  ### References / prior art
  - Sigstore bundle spec: verification material may be a transparency-log entry, a signed timestamp, or both.
  - **`cosign verify-blob-attestation` verifies the same bundles with `--use-signed-timestamps --insecure-ignore-tlog` (+ `--trusted-root`); `gh attestation verify` verifies them natively.**

anthonydahanne:
* Add support for verifying bundles without transparency-log entries, relying on signed RFC 3161 timestamps.

#### Summary
Solves #1251 

#### Release Note
  * Added `VerificationOptions.allowNonTransparencyLogVerification()` (default `false`): when enabled, `KeylessVerifier` verifies bundles that contain **no** transparency-log entry, relying on a signed RFC 3161 timestamp for trusted time. A transparency-log entry that *is* present is still fully verified — the option only relaxes the requirement that one exists.
  * Added a `--allow-non-transparency-log` flag to the `sigstore-cli verify` command exposing the same option.
  * Default behaviour is unchanged: a transparency-log entry remains required unless the option is set.

#### Documentation
Verifying a bundle that has no transparency-log entry (e.g. a GitHub artifact attestation from a **private** repository, signed by GitHub's non-public-good Sigstore instance) now works by opting in:

  * Library: `VerificationOptions.builder().allowNonTransparencyLogVerification(true).build()`, with a trusted root supplied via  `KeylessVerifier.builder().trustedRootProvider(...)`.
  * CLI: `sigstore-cli verify <artifact> --bundle <bundle> --trusted-root <root> --allow-non-transparency-log --certificate-identity <san> --certificate-oidc-issuer <issuer>`.

The bundle must still carry a valid RFC 3161 timestamp, and enabling the option reduces assurance (no transparency-log monitorability), so it should only be used for instances that intentionally do not publish to a log. Covered by the new Javadoc on the option and the CLI `--help`

No separate docs.sigstore.dev change is required, but happy to open one if preferred.


sigstore-java/src/main/java/dev/sigstore/VerificationOptions.java
```diff
@@ -28,6 +29,24 @@ public interface VerificationOptions {
...
+   * <p>Enabling this weakens the guarantees: without a transparency-log entry the signature is not
+   * publicly monitorable/auditable. Only enable it for bundles from instances that intentionally do
+   * not publish to a transparency log (for example, GitHub's Sigstore instance for private
+   * repositories), and pair it with a pinned trusted root. A signed timestamp is still required.
+   *
+   * <p>This only relaxes the <em>requirement</em> that a transparency-log entry exists; a
+   * transparency-log entry that <em>is</em> present in the bundle is still fully verified.
+   */
+  @Default
+  default boolean allowNonTransparencyLogVerification() {
```
loosebazooka: I think to make this more adaptable to other tlog configurations, I would like something that looks more like:
```
  default TLogOptions getTLogOptions() {
    return ImmutableTLogOptions.builder().isEnabled(true).build();
  }

  @Immutable
  interface TLogOptions {
    boolean isEnabled();

    static ImmutableTLogOptions.Builder builder() {
      return ImmutableTLogOptions.builder();
    }
  }
```

sigstore-cli/src/main/java/dev/sigstore/cli/Verify.java
```diff
@@ -113,6 +113,16 @@ static class Policy {
       required = false)
   Path workingDirectory;
 
+  @Option(
+      names = {"--allow-non-transparency-log"},
```
loosebazooka: the CLI doesn't need to be modified yet, unless you're updating the conformance tests at : https://github.com/sigstore/sigstore-conformance

sigstore-java/src/main/java/dev/sigstore/bundle/BundleReader.java
```diff
@@ -50,9 +50,11 @@ static Bundle readBundle(Reader jsonReader) throws BundleParseException {
...
     bundleBuilder.mediaType(protoBundle.getMediaType());
 
-    if (protoBundle.getVerificationMaterial().getTlogEntriesCount() == 0) {
-      throw new BundleParseException("Could not find any tlog entries in bundle json");
-    }
+    // A bundle may legitimately carry no transparency-log entry (for example, GitHub's Sigstore
+    // instance for private repositories, which relies on a signed RFC 3161 timestamp instead).
+    // Such bundles are parsed here; whether their absence is acceptable is a verification-time
+    // policy decision (see VerificationOptions#allowNonTransparencyLogVerification), not a parse
+    // error. Any tlog entries that ARE present must still be well-formed (inclusion proof required).
```
loosebazooka: Can we make this comment a little more concise.
```
A bundle may skip tlog entries in private environments that don't require them.
```

sigstore-java/src/main/java/dev/sigstore/fulcio/client/FulcioVerifier.java
```diff
@@ -150,7 +150,30 @@ private void verifyEmbeddedScts(CertPath certPath) throws FulcioVerificationExce
...
+   * @param signingCertificate the chain to verify
+   * @param allowEmptyCtLogs when {@code true} and the configured trust root declares <em>no</em> CT
+   *     logs, SCT verification is skipped. This is required for Sigstore instances that do not use
+   *     certificate transparency (for example, GitHub's Sigstore instance for private repositories,
+   *     whose trust root carries zero CT logs and whose certificates embed no SCT). When {@code
+   *     false} (the default) a missing CT log is an error, preserving the strict behavior for the
+   *     public-good instance. If the trust root <em>does</em> provide CT logs, SCTs are always
+   *     verified regardless of this flag.
+   */
+  public void verifySigningCertificate(CertPath signingCertificate, boolean allowEmptyCtLogs)
```
loosebazooka: This should again be configured via something like `CTLogOptions` in `VerificationOptions` where the defaults are what public good expects. Any modifications for a private deployment should be explicit.

sigstore-java/src/main/java/dev/sigstore/KeylessVerifier.java
```diff
@@ -132,6 +132,18 @@ public Builder trustedRootProvider(TrustedRootProvider trustedRootProvider) {
...
+      // case below), otherwise fall back to the signing certificate's algorithm. Whether such a
+      // bundle is permitted at all is enforced in verify(...) via
+      // VerificationOptions#allowNonTransparencyLogVerification.
+      if (bundle.getEntries().isEmpty()) {
+        if (bundle.getDsseEnvelope().isPresent()) {
+          return AlgorithmRegistry.HashAlgorithm.SHA2_256;
+        }
+        var publicKey = Certificates.getLeaf(bundle.getCertPath()).getPublicKey();
+        return AlgorithmRegistry.getSigningAlgorithm(publicKey).getHashAlgorithm();
+      }
```
loosebazooka: I think this might be misrepresenting that dsse is always sha2_256. `0.0.1` entries should be handle by the conditional on `0.0.1` everything else can default to the Certificate's publickey mapping to a hashalgorithm. (section starting at line 164) -- we don't need to through the exception at 170 anymore.

sigstore-java/src/main/java/dev/sigstore/KeylessVerifier.java
```diff
@@ -209,7 +222,20 @@ public void verify(byte[] artifactDigest, Bundle bundle, VerificationOptions opt
...
+    // timestamp; verifying those requires opting in and relies on the RFC 3161 timestamp (verified
+    // below) for trusted time.
+    var entries = bundle.getEntries();
+    if (entries.isEmpty() && !options.allowNonTransparencyLogVerification()) {
+      throw new KeylessVerificationException(
+          "Bundle does not contain a transparency log entry. If it is from an instance that does "
+              + "not publish to a transparency log, set "
+              + "VerificationOptions.allowNonTransparencyLogVerification(true) to verify using its "
+              + "signed timestamp instead.");
+    }
```
loosebazooka, line 223: I think we should also disallow ignoring tlog entries that exist, I don't want people accidentally doing something silly (while still aknowledging that we don't have multi tlog support)
```
if (!entries.isEmpty() && options.allowNonTransparencyLogVerification()) {
 // error
}
```

sigstore-java/src/main/java/dev/sigstore/VerificationOptions.java
loosebazooka: Similarly for CTLogOptions

loosebazooka (commented):
I think this mostly looks alright. Would be nice to reduce the verbosity on the code comments.

sigstore-cli/src/main/java/dev/sigstore/cli/Verify.java
anthonydahanne: ok, I'll remove it from this PR

sigstore-java/src/main/java/dev/sigstore/bundle/BundleReader.java
anthonydahanne: yes

sigstore-java/src/main/java/dev/sigstore/fulcio/client/FulcioVerifier.java
anthonydahanne: yes

sigstore-java/src/main/java/dev/sigstore/KeylessVerifier.java
anthonydahanne: ok!

sigstore-java/src/main/java/dev/sigstore/KeylessVerifier.java
anthonydahanne: ok

sigstore-java/src/main/java/dev/sigstore/KeylessVerifier.java
```diff
@@ -132,31 +132,31 @@ public Builder trustedRootProvider(TrustedRootProvider trustedRootProvider) {
...
-        if ("hashedrekord".equals(body.getKind())) {
-          var entry = RekorTypes.getHashedRekordV001(rekorEntry);
-          switch (entry.getData().getHash().getAlgorithm()) {
-            case SHA_256:
-              return AlgorithmRegistry.HashAlgorithm.SHA2_256;
-            case SHA_384:
-              return AlgorithmRegistry.HashAlgorithm.SHA2_384;
-            case SHA_512:
-              return AlgorithmRegistry.HashAlgorithm.SHA2_512;
+      // rekor 0.0.1 hashedrekord entries encode the hash algorithm in the entry itself. Others
```
loosebazooka: comment can be
```
// special case hashedrekord:0.0.1 and dsse:0.0.1 which may contain legacy public key to hash algorithm mappings
```

sigstore-java/src/main/java/dev/sigstore/bundle/BundleReader.java
```diff
@@ -50,9 +50,7 @@ static Bundle readBundle(Reader jsonReader) throws BundleParseException {
 
     bundleBuilder.mediaType(protoBundle.getMediaType());
 
-    if (protoBundle.getVerificationMaterial().getTlogEntriesCount() == 0) {
-      throw new BundleParseException("Could not find any tlog entries in bundle json");
-    }
+    // A bundle may skip tlog entries in private environments that don't require them.
```
loosebazooka: this comment can be removed

sigstore-java/src/main/java/dev/sigstore/fulcio/client/FulcioVerifier.java
```diff
@@ -150,7 +151,19 @@ private void verifyEmbeddedScts(CertPath certPath) throws FulcioVerificationExce
...
+  }
+
+  /**
+   * Verify a signing certificate, applying the given certificate-transparency policy. When SCT
+   * verification is disabled (ex: a private deployment without CT logs) the SCT check is skipped.
+   */
+  public void verifySigningCertificate(CertPath signingCertificate, CTLogOptions ctLogOptions)
+      throws FulcioVerificationException, IOException {
     CertPath fullCertPath = validateCertPath(signingCertificate);
+    if (!ctLogOptions.isEnabled()) {
```
loosebazooka: once again this might allow accidental bypassing CT log checks. This might need to move into verifySct(...), maybe with a method signature change. In there, it should check if sct logs are present before returning and should only return without error if not scts are present, and isEnabled() is false.

sigstore-java/src/main/java/dev/sigstore/bundle/BundleReader.java
anthonydahanne: ok

sigstore-java/src/main/java/dev/sigstore/KeylessVerifier.java
anthonydahanne: ok

sigstore-java/src/main/java/dev/sigstore/fulcio/client/FulcioVerifier.java
anthonydahanne: correct, now it should be properly protected behind `if (ctLogs.isEmpty())`

sigstore-java/src/test/java/dev/sigstore/KeylessVerifierTest.java
```diff
@@ -871,4 +886,133 @@ public void testVerify_wrongArtifactHashAlgorithm() throws Exception {
...
+  // (via `gh attestation trusted-root`) contains zero CT logs. Verifying such a bundle therefore
+  // requires a policy that disables both transparency-log and certificate-transparency checks.
+  //
+  // The zip file contains two files:
+  //   - attestation.sigstore.json : the sigstore bundle (DSSE, in-toto SLSA provenance predicate)
+  //   - trusted_root.jsonl        : the trust roots as JSON Lines. `gh attestation trusted-root`
+  //                                 emits more than one (public-good and GitHub); sigstore-java
+  //                                 consumes a single trust root, so the tests below select the
+  //                                 GitHub one (the line describing the fulcio.githubapp.com CA).
+  private static final String GH_PRIVATE_ATTESTATION_ZIP =
```
loosebazooka: Oh oops, sorry I left a comment on draft in here: we want to avoid zip files in here, I don't think git works efficiently with them. You can create a sub directory in the samples folder if you want which these expanded. Put this gigantic comment instead in a readme next to the bundle, include instructions on how to obtain it.

sigstore-java/src/main/java/dev/sigstore/fulcio/client/FulcioVerifier.java
```diff
@@ -95,7 +96,17 @@ private FulcioVerifier(
 
   @VisibleForTesting
   void verifySct(CertPath fullCertPath) throws FulcioVerificationException {
```
loosebazooka: this method can just go away, we don't need to preserve it.

sigstore-java/src/test/java/dev/sigstore/KeylessVerifierTest.java
```diff
@@ -415,7 +421,9 @@ public void verifyCertificateMatches_noneMatch() throws Exception {
                             .issuer(StringMatcher.string("not-match-again"))
                             .build())));
     Assertions.assertEquals(
-        "No provided certificate identities matched values in certificate: [{issuer:'String: not-match',san:'String: not-match'},{issuer:'String: not-match-again',san:'String: not-match-again'}]",
+        "No provided certificate identities matched values in certificate: [{issuer:'String:"
```
loosebazooka: hrm the formatter seems to have done something funny here.
Can we rearrange this more logically.

loosebazooka (commented):
sorry one final pass I think. But otherwise looks good

sigstore-java/src/test/java/dev/sigstore/KeylessVerifierTest.java
anthonydahanne: gotcha!

sigstore-java/src/main/java/dev/sigstore/fulcio/client/FulcioVerifier.java
anthonydahanne: ok

sigstore-java/src/test/java/dev/sigstore/KeylessVerifierTest.java
anthonydahanne: yes!

anthonydahanne:
hopefully this is good this time 🤞 - some reformatting stuff I don't  really like, but hey, linter is always right 😅

loosebazooka:
Thanks for adding this, and sorry for the slow turnaround.

loosebazooka (approved):

PR #1266, opened 2026-08-11
Handle 409s from rekor

loosebazooka:
If rekor 409s with entry already exists, go grab the entry and provide it to the client.


aaronlew02 (approved):

ghost (commented):
Change to default

loosebazooka:
@marianobayu-sudo  what do you mean?

